PR #87, opened 2014-10-21
Dev/ewma2

dbaarda:
This shows how the EWMA2 filter would look with a dynamic dt using a timer.

The advantage of using dynamic dt is;

1) if the update interval varies because of system load/whatever, then the amount of filtering will not vary.
2) if tracker.cpp's update behavour/interval/whatever changes, then the amount of filtering will not change.

Note this is based on unstable, not your ewma dev branch which seems to include a fair bit of debugging stuff. I don't have an environment setup yet that can make ui changes so the settings are all unchanged. Once I get Qt5 Designer running I'll put together some UI changes for the baseSmoothing, dynamicSmoothing settings we discussed.

### DIFF
--- a/ftnoir_filter_ewma2/ftnoir_filter_ewma2.cpp
+++ b/ftnoir_filter_ewma2/ftnoir_filter_ewma2.cpp
@@ -22,17 +22,9 @@
 // to minSmooth at a rate controlled by the powCurve setting.
 
 
-FTNoIR_Filter::FTNoIR_Filter() :
-    first_run(true),
-    // Currently facetracknoir/tracker.cpp updates every dt=3ms. All
-    // filter alpha values are calculated as alpha=dt/(dt+RC) and
-    // need to be updated when tracker.cpp changes.
-    // TODO(abo): Change this to use a dynamic dt using a timer.
-    // Deltas are smoothed over the last 1/60sec (16ms).
-    delta_alpha(0.003/(0.003 + 0.016)),
-    // Noise is smoothed over the last 60sec.
-    noise_alpha(0.003/(0.003 + 60.0))
+FTNoIR_Filter::FTNoIR_Filter()
 {
+    reset();
 }
 
 void FTNoIR_Filter::receiveSettings()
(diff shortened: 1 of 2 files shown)


sthalik:
Hey @dbaarda do you want to work on it more before the release? You can merge your own pull requests btw.

dbaarda:
I was intending on getting a full dev/run environment working with
qtcreator so I could change the UI and actually test stuff before I sent
push requests, but got distracted by other things (work, life, etc) before
I got it all working.

I am intending on having another crack at it in a couple of months or so,
but probably can't do anything for the next month.

I'll have some changes to ewma2 and kalman when I come back though :-)

On 1 February 2015 at 14:23, Stanisław Halik <notifications@github.com>
wrote:

> Hey @dbaarda <https://github.com/dbaarda> do you want to work on it more
> before the release? You can merge your own pull requests btw.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opentrack/opentrack/pull/87#issuecomment-72349489>.
>



-- 
Donovan Baarda <abo@minkirri.apana.org.au>

sthalik:
@dbaarda to get a measure of noise amount here

<code>        new_delta = target_camera_position[i]-output[i];
        delta[i] = delta_alpha*new_delta + (1.0-delta_alpha)*delta[i];
        // Calculate the current and smoothed noise variance.
        new_noise = delta[i]*delta[i];</code>

an already-smoothed value is used. Perhaps comparing smoothed value to unsmoothed noise is relevant here? An idea here is if delta of smoothed value is large, the user is moving their head and it's not a good measure of noise. But if smoothed delta is small, unsmoothed delta can sensibly represent the actual noise in the process.

dbaarda:
Just following up on your last comment here:

The "delta" is smoothed but with a very small "delta_RC" =1/60 sec time constant. This is to remove very high frequency noise that is too twitchy to possibly be real movements. This (barely) smoothed "delta" is then squared and filtered with a much larger "noise_RC" = 60sec to give the average "noise" variance of "delta" over the last minute. 

Then the current "delta" is compared to the average "noise" to decided what amount of smoothing to use between kMinSmoothing and kMaxSmoothing. The smoothing used is based on the current "delta" scaled between 0~3stddevs using a power curve from kSmoothingScaleCurve. Note kSmoothingScaleCurve=10 is equivalent to sqrt(variance) which is stddev, so the default scaling is linear with stddev.

The idea is the variance over the last minute should automatically give a scale to compare the the current amount of "delta" movement to to figure out the ideal scaling. The variance should represent a normal mixture of noise and real movements, and noise is expected to be smaller than real movements. The current delta should 99.5% fall between 1 and 3 stddevs, with values closer to 3 stddevs being more likely to be real movements (and thus wanting less smoothing for less lag), and values closer to 0 stddevs being more likely to be noise (and thus wanting more smoothing for less jitter).

PR #132, opened 2015-02-14
accela patch

DaMichel:
Turned accela into an exponentially weighted filter with variable rate coefficient. Response is a smooth function of input. Seems to work better for me.


dbaarda:
Note that's what ewma2 is... You might want to take a look at it.

DaMichel:
Yeah but they are broken for me and so i started messing with what works and what i understand.

sthalik:
I can't test this for now since my GPU is broken. Wait some.

sthalik:
@DaMichel can there be a good formula with just one slider? Having two separate settings for sensitivity is just confusing for end users. The hardcoded values present now should be removed definitely though.

DaMichel:
I don't know. Perhaps the dead zone could be adjusted automatically for the amount of noise. But i suppose this is what the Kalman filter is there for ...

I will probably remove the extra slider for translation though because i mostly use the same setting for rotation and translation anyway.

sthalik:
Different games require different translation scales. A suitable value may be "10" on a translation axis but can be "100" the same. That's why there are separate sliders.

sthalik:
Splines are now implemented in Accela.

PR #133, opened 2015-02-14
point tracker patch

DaMichel:
Not so much has come out of my experiments to improve the accuracy of the pointtracker. Probably i'm just tracking the noise really accurately ...  I don't use the hysteresis stuff since i feel it is not accurate enough and noisy during head movements. So i removed that.

Not expecting you to merge this. But since you asked for it ... ;-)


sthalik:
Theoretically meanshift is the right tool for the job here. It's interesting reading this.

Were there visible improvements over the previous setup?

sthalik:
We've tested it and it's less stable in a stationary position.

> com *= 1.0/m;

Often m is zero, then the whole computation gets poisoned by NaNs.

PR #135, opened 2015-02-15
Linear Interpolation of Mapping Curves

DaMichel:
Because i used the curves to amplify the output considerably, the nearest neighbor lookup resulted in jagged movement. This is much better.


sthalik:
Makes sense, thanks.

PR #154, opened 2015-05-05
Fix returning consistent data to get headpose and other minor issues

MartinJG:
Pull request that address the bugs found so far under the hatire plugin issues.


sthalik:
Looks good, thanks

PR #178, opened 2015-07-02
Small improvements to the x-plane plugin

sbabbi:
* Replaced `XPLMRegisterDrawCallback` with ` XPLMRegisterFlightLoopCallback`, so that when the user tries to change the view it does not render a single frame on the new view (it is still not possible to change the view while the plugin is running, at least on Xplane 10.35).
* Added a command (*opentrack/toggle*) to disable the plugin from inside the sim.
* Added a command (*opentrack/toggle_translation*) to disable only the traslational components, so that we can still move around if OT only provides the rotations.
* Reinitialize the translational offsets when we re-enable the plugin, so that we can disable the plugin, move around (a long distance), and then re-enable the plugin in the new position.


sthalik:
Thanks!

PR #325, opened 2016-03-10
initial wiimote tracker

fred41:
based on hid-wiimote kernel module, 
works already stable and fast here,
usage instructions in doc/readme.txt and ui 'about' box

TODO: 
userinterface adaptions/improvements
considering/designing an more asynchron approach for opentrack generally,
currrent synchron polling is not fully sufficient, imo


sthalik:
- `select(2)` shouldn't wait for one nanosecond. Make it wait for a few seconds.
- `select(2)` should close and stop the loop unless in special case of `EINTR`
- Remove the rest of PT tracker stub and leave two classes, tracker and worker thread. By worker thread I mean `QThread`, not `pthread`. There's plenty of functions that only made sense for the PT tracker but they were left there.
- Don't use `new someclass`, use `std::shared_ptr<t>` and `std::make_shared<t>(x)`.
- Remove the widget and don't set parent frame's layout if you don't use it.

fred41:
First, thanks for your hints, let's discuss it point by point.
* `select(2)` was actually meant to be nonblocking (1 usec), but i have to agree, it makes more sense to wait in select and wait for new events, than in the while loop, changed to blocking select
* the blocking approach makes the loop needless, see above, changed
* the `ir_cache` error checking is fully sufficient, i think. Please explain, if you see a problem.

In respect to PT tracker and dependicies/duplicate code, i think for now, it is suitable to keep it isolated per plugin. There are new algorithms available meanwhile, that allow higher precision together with similiar or better performance. By keeping the point tracker code isolated per tracker, it is easier to test improvements without affecting other plugins functionality. And last but not least, a device like WiiMote, has some special propertys in respect to precision and timing, that ordinary web cams dont share.

FOV: can be always calculated from focal length (knowing sensor resolution) so saving it as additional number would be redundant, actually.

What do you mean with:
'Point 3D coordinates are known and there's no need to specify them, similarly.' ?

sthalik:
> What do you mean with:
> 'Point 3D coordinates are known and there's no need to specify them, similarly.' ?

The device has a known layout and 3D coordinates of the three IR points can be hardcoded. Wiimote is a single design, each copy is identical.

> the ir_cache error checking is fully sufficient, i think. Please explain, if you see a problem.

I see, `INVALID_EV_VALUE` is defined above image width and height.

> In respect to PT tracker and dependicies/duplicate code, i think for now, it is suitable to keep it isolated per plugin.

There's hundreds of lines of lines of duplicated code. `point_tracker.cpp` looks verbatim and is 268 lines alone.

If you want to pass review, isolate the following subsets of the code:

- camera returning an all-black image of wiimote's frame size
- point extraction as a member function
- a dialog with Wiimote-specific options, if any. Only the following values from PT also make sense for Wiimote:
- - value\<bool\> dynamic_pose;
- - value\<int\> init_phase_timeout;
- - model centroid if Wiimote is to be mounted on a hat or headphones
- - point coordinates that'll be used to initialize `PointModel` in PT
- - focal length function

Basing on that, original PT code can be abstracted to work both with LEDs and Wiimote. I'm willing to do the work. I can put it on a branch and merge once we're satisfied with both PT and Wiimote working from the same codebase.

> By keeping the point tracker code isolated per tracker, it is easier to test improvements without affecting other plugins functionality.

I won't merge a couple hundred lines of copy-pasted code with up to a hundred lines changed. Also there are no changes for the most part, only for the main loop and camera class.

> fov can always be calculated from focal length

That's interesting, looks like original PT doesn't divide focal length by image width like it should. I'll look into it.

fred41:
A WiiMote doesnt return images, it returns points. There is no need to extract anything, sorry.

Stanislaw, please don't missunderstand me, but meanwhile i think i should implement my ideas in a little independent linux/wiimote only project. Please reject my pull request.

I really appreciate your commitment and your work for opentrack. Have a good time.

sthalik:
I'm rejecting as per your request

PR #372, opened 2016-06-12
Update README.md

MathijsG:
'Binary releases' sounds way too abstract for people who aren't technical and just want to download this software. So they would rather search for "download"


sthalik:
Thanks

PR #383, opened 2016-07-08
Kalman filter

DaMichel:
Meanwhile I attempted to fix the Kalman filter.

It turned out to be an issue with the timer. Possibly, the return value was assumed to be of a different unit than it really is. 

Rather than fixing the use of the QT timer, I instead replaced it with your `Timer` class for consistency.

Tbh. I'm a bit disappointed by the praised Kalman filter. It reacts either extremely slowly or it overshoots and comes back close to the real position in a very jittery way. That makes me want to try to make the process noise dependent on the predicted movement speed ... 

Also pinging @dbaarda

### DIFF
--- a/filter-kalman/CMakeLists.txt
+++ b/filter-kalman/CMakeLists.txt
@@ -1,8 +1,6 @@
-if(FALSE)
 find_package(OpenCV 3.0 QUIET)
 if(OpenCV_FOUND)
     opentrack_boilerplate(opentrack-filter-kalman)
     target_link_libraries(opentrack-filter-kalman ${OpenCV_LIBS})
     target_include_directories(opentrack-filter-kalman SYSTEM PUBLIC ${OpenCV_INCLUDE_DIRS})
-endif()
 endif()
--- a/filter-kalman/ftnoir_filter_kalman.h
+++ b/filter-kalman/ftnoir_filter_kalman.h
@@ -18,6 +18,7 @@
 #include <QWidget>
 #include "opentrack-compat/options.hpp"
 using namespace options;
+#include "opentrack-compat/timer.hpp"
 
 struct settings : opts {
     value<int> noise_stddev_slider;
@@ -39,7 +40,8 @@ public:
     static constexpr double accel_stddev = (accel*4/(dt_*dt_))/3.0;
     cv::KalmanFilter kalman;
     double last_input[6];
-    QElapsedTimer timer;
+    Timer timer;
+    bool first_run;
     settings s;
     int prev_slider_pos;
 };
--- a/filter-kalman/kalman.cpp
+++ b/filter-kalman/kalman.cpp
@@ -66,7 +66,7 @@ void FTNoIR_Filter::reset() {
     for (int i = 0; i < 6; i++) {
         last_input[i] = 0;
     }
-    timer.invalidate();
+    first_run = true;
 }
 
 void FTNoIR_Filter::filter(const double* input, double *output)
@@ -76,16 +76,24 @@ void FTNoIR_Filter::filter(const double* input, double *output)
         reset();
         prev_slider_pos = s.noise_stddev_slider;
     }
-    // Start the timer if it's not running.
-    if (!timer.isValid())
+    // Start the timer on first filter evaluation.
+    if (first_run)
+    {
         timer.start();
+        first_run = false;
+        return;
+    }
+
     // Get the time in seconds since last run and restart the timer.
-    const double dt = timer.restart() / 1000.;
+    const double dt = timer.elapsed_seconds();
+    timer.start();
+    
     // Note this is a terrible way to detect when there is a new
     // frame of tracker input, but it is the best we have.
     bool new_input = false;
     for (int i = 0; i < 6 && !new_input; i++)
         new_input = (input[i] != last_input[i]);
+
     // Update the transitionMatrix and processNoiseCov for dt.
     double accel_variance = accel_stddev * accel_stddev;
     double a = dt * dt * accel_variance;  // dt^2 * accel_variance.


sthalik:
I don't like it either. KF is used here to get the "real" pose but that's not what filters are for. Filters are meant to return a pose very close to the last one. Human head shakes on its own. Try setting up Accela for a handheld clip compared to a worn clip.

DaMichel:
I would be happy with an accurate replication of just the wobble of my head. However, my head wobble is completely overshadowed by the noisy measurement from my camera.
Your Accela filter does a very good job. I find it totally sufficient. I was messing with the Kalman filter for my own educational purposes :-)

DaMichel:
I made simple sim written in python to help with analysis of parameters and model changes. It is in my repo https://github.com/DaMichel/opentrack/commit/b309406f206a6371822c470fb29903f725b936f4 Perhaps it will be useful for you or anybody who wants to play with KF some more.

sthalik:
@DaMichel is it possible to add change delta as an error measure to KF? That is, more delta, worse it is.

DaMichel:
Well I guess. But a systematic approach with proven methods would be in order. Adaptive KF is kind of what you propose. AKF is designed to address our problem that is state estimation of systems that exhibit sudden changes that interrupt relatively steady motion.

sthalik:
@DaMichel KF addresses the "real" data readout wrt gaussian noise dist. But opentrack filters address the need for idempotence while caring a little bit about the input readout.

dbaarda:
A quick summary of Kalman, is a weighted average between the measured
inputs and a model's predictions, weighted by their error variance. In
theory it can filter with less lag by using the model to
effectively "predict the future", instead of just taking an average of the
past measurements. In practice, how well it works depends on how good the
model is at predicting the future, and now good your estimates of the error
variance is.

The current implementation has a very simple model that assumes constant
velocity in each of the dimensions/angles. The process noise covariance
matrix for the model errors assumes the "model error variance" is in the
form of accelerations of a given variance. The measurement noise covariance
matrix for the measurement errors assumes each axis has equal error
variance set by the noise_stddev slider and are completely uncoupled.

First, this model is pretty crap for tracking head movements, which don't
tend to have constant velocity, but rather jump between fixed positions
over a limited range. For face tracking, there is also coupling between
axies; turning and tilting your head also moves the "between the eyes
point" left/right and up/down.

Second, the covariance matrixes are very rough guesses. The process noise
covariance matrix for the simple model is not too bad, but does assume the
acceleration variance is identical for all axies... ie it assumes we
"accelerate" our head the same amount in every direction, even rotation.
Since this error variance is used for a weighted average, this doesn't
matter that much, what matters is the relative magnitudes for these
compared to the measurement noise covariance matrix magnitudes. However,
the measurement noise covariance matrix assumes the measurement noise is
also identical and uncoupled in all dimensions. I'm betting the trackers
have quite different noise levels for different dimensions (at least angles
vs positions), and their errors also have some degree of coupling between
at least some axies.

The behaviour of this simple model when noise_stddev is set high is to
trust the constant velocity model more than the measurements. This means it
will tend to lag when head movement starts, and overshoot when head
movement stops. When noise_stddev is set low, it will trust the
measurements more, tending towards the raw noisy unfiltered inputs.

I think what Stanislaw is proposing is to make the noise_stddev setting
vary based on the error between the filtered output and the raw
measurement, the same way is done in the EWMA filter (and accella?). This
would make the measurement noise covariance matrix vary, adjusting how much
the model vs measurements are trusted based on how much error there is
between filtered vs raw values. This might improve it a fair bit, but
probably for the wrong reasons... it will tend to abandon the model when
the head "jerks" to a new position, and trust it when the head stops... but
then you might as well have a model that assumes zero velocity... which BTW
ends up being identical to EWMA.

A far better fix for kalman would be to come up with a better model and/or
covariance matrixes. I wanted to come up with a model for face tracking
that assumed the tracked "middle of the eyes" point moved on a head that
twisted/tilted around a hinge point "neck" at the rear bottom of the head.
This would end up being an extended kalman filter because it ends up being
non-linear, so the transitionMatrix needs to be updated every frame too.

However, I don't think I'll ever find time to do that.



On 10 July 2016 at 19:26, Stanisław Halik <notifications@github.com> wrote:

> @DaMichel <https://github.com/DaMichel> KF addresses the "real" data
> readout wrt gaussian noise dist. But opentrack filters address the need for
> idempotence while caring a little bit about the input readout.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/opentrack/opentrack/pull/383#issuecomment-231579483>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AB85sw51QRCH2tT1AYR5XHqfnXSt59wOks5qULrcgaJpZM4JIDU3>
> .
>



-- 
Donovan Baarda <abo@minkirri.apana.org.au>

sthalik:
@dbaarda I'm proposing adding deltas as another separate dimension, then supplying all zeros for them each correction iteration. Weight these fake deltas enough so that Kalman considers no movement better than minor movement.

sh

DaMichel:
@sthalik I guess i misunderstood you earlier. So using deltas as velocity "measurement" but with thresholding so that measurement returns zero if the actual delta is too small.

@dbaarda I agree for the most part. But I think Kalman Filter can never properly model head movements because of the assumed Gaussian Noise. Therefore some tricks are needed, i.e. adapting the covariance matrices dynamically to the current situation.
I think your proposal for an improved head model is orthogonal to this.

sthalik:
@DaMichel maybe even always expect zero deltas to make the filter weight idempotence vs absolute error. If that's possible at all.

DaMichel:
@sthalik: Finally uploaded my experimental code to a new repository.
https://github.com/DaMichel/opentrack-prototyping
Especially the AKF2 variant does well enough for my taste. 
https://github.com/DaMichel/opentrack-prototyping/blob/master/kalman_simulation2.py#L149
Considering that even standard KF does quite well I'm thinking that it is worth to take a closer look at the KF in opentrack again.

On a side note: I had trouble with using multiple prediction steps in-between measurement updates because whenever a measurement update came the trajectory "snapped" to the measurement point resulting in a visible discontinuity. I think there is no reason to believe that it should not be the case, @dbaarda (?!)

sthalik:
@DaMichel want a repo in the project?

Accela is using rigid cubic interp (Catmull-Rom).

If trajectory snaps it means (with my limited understanding) that KF overvalues input accuracy over its model too much. Maybe keep "surenesses" the same with multiple predictions like that.

sthalik:
@DaMichel add yourself to readme.md as a contributor, please.

sthalik:
Should we use Eigen or opencv for the matrix ops? I assume adding stuff to simple-mat is overkill.

DaMichel:
@sthalik Regarding first question. Idk. Is it possible for you to clone my repo and put it into the project whenever you feel like it?

Secondly. Regarding trajectory "snapping". I was thinking about running KF updates at rate of PT, i.e. without extra time update. Then compute final filter output from interpolation of position estimates.

Will add myself to readme.

Eigen is great. However we already use opencv for KF and it has everything we need. So this is

sthalik:
@DaMichel my activity is limited at the moment. I'd like you to implement it. Failing that, I'll implement it with a delay.

It's not a problem to add Eigen, especially since the build machine can have everything needed. Furthermore, opencv is build statically now and it really adds up into the time producing the .dll. In my opinion, less modules requiring big static dependencies the better. Eigen is entirely header-only, is that correct?

We won't gain on binary size either since opencv is linked statically into each .dll that needs it anyway.

sthalik:
@DaMichel also, running more frequent updates looks better as the ingame end result.

dbaarda:
The extra predicts between updates was intentional because in opentrack the
filter update rate is independent of the input update rate. You shouldn't
update with stale input data because it's a lie that just adds lag with a
false decrease in the accumulated error variance. However, you can do extra
predicts between input updates to give you incremental output estimates
from the model at a faster rate than the input updates. When you do this
the accumulated error variance increases with each prediction from the
model's added error, so when an input update does come it is trusted even
more than the model.

That it "snaps" is a clear indication that the model is not good. If the
model was good, then it's incremental predictions should end up pretty
close to the correct position when the next input arrives, so there should
be little difference to "snap" to.

If you increase the input noise variance it will decrease the "snap" effect
because it will trust the inputs less and will not immediately "snap" to
them. However, it does mean it trusts the model more, which the snapping
already tells us is bad, so although it will be smoother, it will lag and
overshoot more.

Another option would be to not do incremental predictions at all, which
would pin the output update rate to the input rate.

On 24 Jul 2016 6:55 PM, "DaMichel" <notifications@github.com> wrote:

@sthalik <https://github.com/sthalik>: Finally uploaded my experimental
code to a new repository.
https://github.com/DaMichel/opentrack-prototyping
Especially the AKF2 variant does well enough for my taste.
https://github.com/DaMichel/opentrack-prototyping/blob/master/kalman_simulation2.py#L149
Considering that even standard KF does quite well I'm thinking that it is
worth to take a closer look at the KF in opentrack again.

On a side note: I had trouble with using multiple prediction steps
in-between measurement updates because whenever a measurement update came
the trajectory "snapped" to the measurement point resulting in a visible
discontinuity. I think there is no reason to believe that it should not be
the case, @dbaarda <https://github.com/dbaarda> (?!)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/opentrack/opentrack/pull/383#issuecomment-234770748>,
or mute the thread

sthalik:
@dbaarda 
> Another option would be to not do incremental predictions at all, which would pin the output update rate to the input rate.

Then the only solution is being one frame late. Interpolate between the last-1 and last frame using last dt.

sthalik:
@DaMichel see how current KF code checks for fresh input. It's not pretty but prevents the aforementioned issue.

@dbaarda you can't get a good model anyway without overfitting. Head center decouples rotation from translation. What can you do?

DaMichel:
> It's not a problem to add Eigen, especially since the build machine can have everything needed. Furthermore, opencv is build statically now and it really adds up into the time producing the .dll. In my opinion, less modules requiring big static dependencies the better. Eigen is entirely header-only, is that correct?
Eigen is header-only. At least the core things that I happened to need. Both Eigen and OpenCV have their merits, like we already have the OpenCV based codes. But I don't mind using either and will accept your suggestion.

dbaarda:
Kalman with a bad model is worse than not using kalman at all. A bad model
will inject more error into the output, adding to the input noise error
instead of cancelling it out.

There are a bunch of simple model improvements that might make it better,
including things like;

1) Assume constant acceleration instead of constant velocity. This requires
extending the state space to include acceleration for each dimension so you
have 18x18 matrices for the model. Alternatively you could calculate
acceleration separately and feed it into the model as a demand.

2) Assume the head prefers to be still, and have the model decay velocity
over time. This involves changing the state prediction matrix to have
T/(T+dt) terms on the diagonal for the velocity terms, where T is the time
for velocity to halve. This will tend to increase the lag and snap effects
when the head starts moving, but will reduce the overshoot at the end.

3) same assumption as 2), but instead of decay velocity, assume constant
acceleration against direction of velocity. Implement this as a model
demand calculated from velocity. Probably similar affects as 2), but maybe
less for fast head movements and more for slow head movements.

4) Assume the head prefers to be centered. Implement as decay of position
with T/(T+dt) terms on diagonal for position, where T Is the time for the
head to return half way to centered. This will have reduced lag, snap
effects when at or returning head to centered, and increased when turned or
turning away from centered.

5) Same assumption as 4), but implement as velocity decaying towards a
value towards the  center, proportional to position. Implement as
-K*dt/(T+dt) matrix values where position affects velocity, and T/(T+dt)
terms on the velocity diagonals, where -K*position is the desired velocity
and T is the time to decay half way to it. Probably similar effects as 4),
but maybe works better.

6) Some fancy heuristic about head movements as a function of position,
maybe assuming limited range of motion and velocities. Probably best
implemented as a heuristic function calculating demands for the model from
state positions and velocities.

7) Some combination of the above.

I would probably collect raw position and velocity data to try and identify
how the head actually moves. I suspect that it tends to snap between static
positions with either constant acceleration or constant velocity when
moving. The range and velocity characteristics are probably different for
different axies. I would also be quite surprised if even with "head center"
there is no coupling of axies.

I also suspect that lag/snap/errors are subjectively worse when the head
stops moving than when it starts moving or during movement. Hence changes
like the decaying velocity that reduces overshoot at the expense of
increased lag and jitter when moving will work OK.

sthalik:
Are there some superior filtering methods for our purpose of not caring about accuracy, aka with gaussian noise removal? We don't care about the accuracy in opentrack filtering. We care about not doing small changes while still having the output remotely resemble the input.

Is there an online filter minimizing both abs differences and deltas according to weights?

DaMichel:
I learned about particle filters. If i understand correctly they would allow for a non-Gaussian model for head movements. Like we could assume an acceleration probability density that has a peak at zero superimposed on a uniform non-zero "background" or something like that.

sthalik:
@DaMichel but particle filters are meant to estimate the "true position" also. If we got the 100% true position, the view would jump around ingame. We solve a different thing - idempotent values with varying responses being pleasant for the user.

PR #398, opened 2016-07-29
Track Logging

DaMichel:
This is probably worthy of a pull request. I hope the capability to log pose data will be useful not just for me.

Btw., I replaced my previous branch of the same name due to the rebase.

### DIFF
--- a/gui/options-dialog.cpp
+++ b/gui/options-dialog.cpp
@@ -11,6 +11,7 @@
 #include <QPushButton>
 #include <QLayout>
 #include <QDialog>
+#include <QFileDialog>
 
 static QString kopts_to_string(const key_opts& kopts)
 {
@@ -72,6 +73,9 @@ OptionsDialog::OptionsDialog(std::function<void(bool)> pause_keybindings) :
 
     tie_setting(main.center_method, ui.center_method);
 
+    tie_setting(main.tracklogging_enabled, ui.tracklogging_enabled);
+    tie_setting(main.tracklogging_filename, ui.tracklogging_filenameedit);
+
     struct tmp
     {
         key_opts& opt;
@@ -102,6 +106,8 @@ OptionsDialog::OptionsDialog(std::function<void(bool)> pause_keybindings) :
             connect(val.button, &QPushButton::pressed, this, [=]() -> void { bind_key(val.opt, val.label); });
         }
     }
+
+    connect(ui.tracklogging_fileselectbtn, SIGNAL(clicked()), this, SLOT(browse_datalogging_file()));
 }
 
 void OptionsDialog::bind_key(key_opts& kopts, QLabel* label)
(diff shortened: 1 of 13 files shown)


sthalik:
I like the implementation. :+1: 

Please remove all exception usage however since release builds won't compile otherwise.

gui/options-dialog.cpp
```diff
@@ -163,3 +169,23 @@ void OptionsDialog::doCancel()
     close();
 }
 
+void OptionsDialog::browse_datalogging_file()
+{
+    QString filename = ui.tracklogging_filenameedit->text();
+    if (filename.isEmpty())
+        filename = QDir::currentPath();
+    /* Sometimes this function freezes the app before opening the dialog.
```
DaMichel: @sthalik well there is this little issue though. This should really be idiot proof but it froze twice already.
I will continue using it with the non-native dialog with which it might work.

sthalik: Remove usages of `CoInitializeEx` from the tree and it should be fine. If this works I'll fix it in the UI thread.

DaMichel:
> Please remove all exception usage however since release builds won't compile otherwise
There should be no exception thrown unless I missed something. I actually tested the case where the file cannot be written, if this is what you have in mind.

sthalik:
@DaMichel there cannot be any references to the C++ exception machinery or RTTI at all. No matter if it's reachable, the compiler will barf seeing a `try` or a `throw` regardless.

sthalik:
@DaMichel try building with GNU using `-fno-exceptions -fno-rtti`.

DaMichel:
What references are you talking about?

sthalik:
@DaMichel don't use identifiers `try` and `throw` at all in the code.

DaMichel:
I don't. Not in the last commit. You want me to rebase everything into a single commit so that the previous code that used exceptions is removed from the repo?

sthalik:
@DaMichel no. Merge it.

sthalik:
To try fix the dialog, in gui/main.cpp, after

<pre>int main(int argc, char** argv)
{
#ifdef _WIN32</pre>

add

<pre>{
    const HRESULT hr = CoInitializeEx(NULL, COINIT_MULTITHREADED);
    if (FAILED(hr))
        qDebug() << "failed CoInitializeEx" << hr << GetLastError();
}</pre>

Tell me if it helps with native dialog.

sthalik:
Also try change in the same file when building with MSVC++

<pre>// on MSVC crashes in atexit
#ifdef _MSC_VER
</pre>

to

<pre>// on MSVC crashes in atexit
#if 0
</pre>

and see if you get crash on exit.

DaMichel:
> and see if you get crash on exit.
Apparently not. At least not in debug build.

DaMichel:
> add
> 
> {
>     const HRESULT hr = CoInitializeEx(NULL, COINIT_MULTITHREADED);
>     if (FAILED(hr))
>         qDebug() << "failed CoInitializeEx" << hr << GetLastError();
> }

I actually removed `CoInitializeEx` everywhere. Things still work. I will do the merging now without the dialog related fix attempts though.

PR #400, opened 2016-07-31
kalman filter revised

DaMichel:
So I was playing around for a bit ... (too long) and realized my idea for adaptive KF and a small model change. Moreover I added a deadzone filter that adapts to the predicted state variance. The results is quite usable IMO.

I don't think it is possible to do much better with KF including the things that @dbaarda suggested. I tried. 

Anyway, I also removed the opencv dependency. It is now written for eigen3.


filter-kalman/kalman.cpp
```diff
@@ -84,54 +223,72 @@ void FTNoIR_Filter::filter(const double* input, double *output)
...
+    
+    {
+        // Compute deadzone size base on estimated state variance.
+        // Given a constant input plus noise, KF should converge to the true (constant) input.
+        // This works indeed. That is the output pose becomes very still afte some time.
+        // At this point the estimated cov should be minimal. We can use this to
+        // calculate the size of the deadzone, so that in the stationary state the
+        // deadzone size is zero. Thus the tracking error due to the dz-filter 
+        // becomes zero.
+        PoseVector variance = kf.state_cov.diagonal().head(6);
```
DaMichel: I should add that this only works for the adaptive KF where the discrepancy between measured position and predicted position causes the process noise cov to increase.
In normal KF state cov is constant.

sthalik:
@DaMichel I fixed the cmake machinery and added align ctors. See <<https://github.com/opentrack/opentrack/tree/kalman-test>>. Previously it crashed on startup and debug compile wouldn't work.

Are you fine with my changes? You can do `git reset --hard` to the changes, do stuff, then force push to your own branch.

sthalik:
@DaMichel please add your copyright to the top of the files. I don't reserve any rights to the files.

filter-kalman/ftnoir_filter_kalman.h
```diff
@@ -10,40 +10,115 @@
...
+    static constexpr double adaptivity_window_length = 0.5; // seconds
+    static constexpr double deadzone_scale = 2.;
+    static constexpr double deadzone_exponent = 4.0;
+    // these values worked best for me (MW) when taken with acompanying measured noise stddev of ca 0.1 (rot) and 0.01 (pos). 
+    static constexpr double process_sigma_pos = 0.05;
+    static constexpr double process_simga_rot = 0.5;
+
+    static double map_slider_value(int v)
+    {
+        return std::pow(10., v * 0.04 - 3.);
```
sthalik: This scaling's weird from my usage. Can we please make it linear? The ui slider can have larger resolution with higher single/page step value.

sthalik:
@DaMichel is there a reason for using heap-allocated matrices when the size is known due to constexpr?

filter-kalman/ftnoir_filter_kalman.h
sthalik: At least maybe use `4` rather than `10` as the exponentiated value?

filter-kalman/kalman.cpp
```diff
@@ -9,73 +9,212 @@
...
+    //qDebug() << "alpha = " << alpha;
+}
+
+
+PoseVector DeadzoneFilter::filter(const PoseVector &input)
+{
+    PoseVector out;
+    for (int i = 0; i < input.rows(); ++i)
+    {
+        const double dz = dz_size[i];
```
sthalik: Maybe add some small epsilon like `1e-8` to the deadzone? It looks easy for it to be zero. We flush denormals btw.

filter-kalman/ftnoir_filter_kalman.h
DaMichel: Okay, will take closer look at sliders.

DaMichel:
> is there a reason for using heap-allocated matrices when the size is known due to constexpr?

To not fill the stack with large object ...  Now that I think of it, ultimately most of these get heap allocated either way. Will change them to fixed size.

filter-kalman/kalman.cpp
DaMichel: Yes. would be better.

sthalik:
@DaMichel can I merge this and then my branch with fixes?

DaMichel:
I just played your changes on top of mine. Will update this pull request soon.

DaMichel:
Regarding the slider. I really like the high dynamic range one gets with log scaling.
Otherwise it gets hard if not impossible to pick small values with any degree of accuracy.

The nicest I could come up with in reasonable time is something like
```
    static double map_slider_value(const slider_value &v_)
    {
        double v = v_;
        //return std::pow(10., v * 4. - 3.);
        if (v < 0.)
            return 0.001;
        else if (v < 0.25)
        {
            v = v / 0.25;
            v = int(v * 90. + 10. + 0.5) * 0.01;
            return v * 0.01;
        }
        else if (v < 0.5)
        {
            v = (v - 0.25) / 0.25;
            v = int(v * 90. + 10. + 0.5) * 0.01;
            return v * 0.1;
        }
        else if (v < 0.75)
        {
            v = (v - 0.5) / 0.25;
            v = int(v * 90. + 10. + 0.5) * 0.01;
            return v * 1.;
        }
        else if (v < 1.0)
        {
            v = (v - 0.75) / 0.25;
            v = int(v * 90. + 10. + 0.5) * 0.01;
            return v * 10.;
        }
        else
            return 10.;
    }
```
which uses equidistant steps between powers of 10.

sthalik:
Since the original slider range was 100, how about:

`return std::pow(10., slider.cur() * 4 - 3);`

You can make the range 0->100 in the slider_value ctor however and not need any manual mapping. That's why there's a separate slider type.

DaMichel:
I would like to keep the kind of number display as generated by my code above. Now I pushed a generalized version of it. The nice thing about it is that it changes only one digit at a time from step to step.

I think other outstanding issues are taken care of. Copyright is changed. And I tweaked the dead zone a bit more and clarified the comment. There is now always a bit of dead zone, depending on the noise settings. It should never go down to zero.

sthalik:
In header just add year to your copyright and :+1: merge.

DaMichel:
Great. Thanks.

sthalik:
Add yourself to contributors please.

DaMichel:
In the readme?

sthalik:
Yes.

PR #421, opened 2016-08-20
Update ftnoir_tracker_udp.cpp

sheepsskullcity:
On success, bind() function returns true. So while(!should_quit) loop was always skipped.


sthalik:
That makes sense, thank you.

PR #428, opened 2016-08-23
PT: improving point extractor's precision

DaMichel:
I finally finished implementing additional mean-shift filtering of the tracked blob-centers.

More importantly, I'm now rather sure that it is actually good. See comparison video on dropbox https://www.dropbox.com/sh/38get6zoksw7bw4/AAAoHiGetDZn_KFHLbHLKV3Ra?dl=0

Moreover, I guess I went a little bit over board trying different methods to improve the tracker's precision. To this end I implemented some prototypes in python. It is probably not immediately useful. But if you want to toy around with things some day, I guess this might be a solid basis.
https://github.com/DaMichel/opentrack-prototyping/blob/master/pointextractor.py

### DIFF
--- a/tracker-pt/point_extractor.cpp
+++ b/tracker-pt/point_extractor.cpp
@@ -7,6 +7,7 @@
  */
 
 #include "point_extractor.h"
+#include "point_tracker.h"
 #include <QDebug>
 
 #ifdef DEBUG_EXTRACTION
@@ -18,6 +19,62 @@
 #include <cmath>
 #include <algorithm>
 #include <cinttypes>
+
+
+
+/*
+http://en.wikipedia.org/wiki/Mean-shift
+In this application the idea, is to eliminate any bias of the point estimate 
+which is introduced by the rather arbitrary thresholded area. One must recognize
+that the thresholded area can only move in one pixel increments since it is
+binary. Thus, its center of mass might make "jumps" as pixels are added/removed
+from the thresholded area.
+With mean-shift, a moving "window" or kernel is multiplied with the gray-scale
+image, and the COM is calculated of the result. This is iterated where the
+kernel center is set the previously computed COM. Thus, peaks in the image intensity
+distribution "pull" the kernel towards themselves. Eventually it stops moving, i.e.
+then the computed COM coincides with the kernel center. We hope that the 
+corresponding location is a good candidate for the extracted point.
+The idea similar to the window scaling suggested in  Berglund et al. "Fast, bias-free 
+algorithm for tracking single particles with variable size and shape." (2008).
+*/
+static cv::Vec2d MeanShiftIteration(const cv::Mat &frame_gray, const cv::Vec2d &current_center, double filter_width)
+{
+    // Most amazingling this function runs faster with doubles than with floats.
+    const double s = 1.0 / filter_width;
+
+    auto EvalFilter = [&current_center, s](int row, int col)
+    {
+        double dx = (col - current_center[0])*s;
+        double dy = (row - current_center[1])*s;
+        double f = std::max(0.0, 1.0 - dx*dx - dy*dy);
+        return f;
+    };
+
+    double m = 0;
+    cv::Vec2d com(0.0, 0.0);
+    for (int i = 0; i < frame_gray.rows; i++)
+    {
+        auto frame_ptr = (uint8_t *)frame_gray.ptr(i);
+        for (int j = 0; j < frame_gray.cols; j++)
+        {
+            double val = frame_ptr[j];
+            val = val * val; // taking the square wights brighter parts of the image stronger.
+            val *= EvalFilter(i, j);
+            m += val;
+            com[0] += j * val;
+            com[1] += i * val;
+        }
+    }
+    if (m > 0.1)
+    {
+        com *= 1.0 / m;
+        return com;
+    }
+    else
+        return current_center;
+}
+
 
 PointExtractor::PointExtractor()
 {
@@ -136,7 +193,7 @@ void PointExtractor::extract_points(cv::Mat& frame, std::vector<PointExtractor::
                 if (radius > region_size_max || radius < region_size_min)
                     continue;
                 const double norm = double(m00);
-                blob b(radius, cv::Vec2d(m10 / norm, m01 / norm), vals/sqrt(double(cnt)));
+                blob b(radius, cv::Vec2d(m10 / norm, m01 / norm), vals/sqrt(double(cnt)), rect);
                 blobs.push_back(b);
                 {
                     char buf[64];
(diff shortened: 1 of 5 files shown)


tracker-pt/point_extractor.cpp
```diff
@@ -19,6 +19,108 @@
...
+    brightness found surrounding the central blob.
+
+    The function should be applied to an ROI surrounding the blob under consideration.
+
+    binary_inv indicates blobs by values equal to *0*!
+*/
+static void RemoveNearbyBlobs(cv::Mat &frame_gray, const cv::Mat &binary_inv, const cv::Vec2i &blob_center)
+{
+    cv::Mat distance(frame_gray.size(), CV_32FC1);
+    cv::Mat labels(frame_gray.size(), CV_32SC1);
```
sthalik: Allocate only once. malloc/free takes a lot of cycles with 75/187 Hz.

tracker-pt/point_extractor.cpp
```diff
@@ -19,6 +19,108 @@
...
+}
+
+/*  The mean-shift iteration is sensitive to nearby blobs, not belonging to the central blob the
+    position of which we want to find. This function attempts to remove the intensity peak pertaining
+    to such nearby blobs.
+
+    It computes a discrete voronoi diagram based on the distances from thresholded blobs (binary_inv).
+    Then it fills the image area of voronoi regions pertaining to other blobs with the average 
+    brightness found surrounding the central blob.
+
```
sthalik: The sheer complexity, and 1.2 megs per each float matrix like these below.

How much CPU usage at 187 Hz in 320x240? How much CPU usage at 75 Hz in 640x480? What CPU?

sthalik:
This convolves each pixel, voronoi is expensive, uses lots of memory. I'd like to talk about it and try find some less computationally-expensive way to do the same.

How much of the output noise exists due to bad point extraction and how much due to user's head being the controller? Find a working filter setup with the clip held in a non-shaking hand then wear the clip. My own conclusions were that it's the head that's imprecise.

Also, if some points move toward converging then it shouldn't affect output idempotence by much, especially if not moving. The fundamental confusion here is that in truth the result needs be only remotely accurate while idempotence needs preserving.

I recommend you generate a sequence of frames from perspective transform of the clip, rotating it very little. Then do sum of squares of deltas (matrix norm, euler angles) from coplanar POSIT as applied  to the frame sequence.

tracker-pt/point_extractor.cpp
DaMichel: I don't think it is that bad. Beside, PT runs at the frame rate of the camera, does it?

tracker-pt/point_extractor.cpp
DaMichel: I didn't check how fast it is. It still ran at 30 fps. But I have a few cores to spare so ... I'm going to time it, though.

DaMichel:
Well in the videos I wore my tracking cap, obviously. So I guess I can keep my head quite still because there is a noticeable difference between my patch and original. IMO less measurement noise, not to be mistaken for head wobble noise, is good because it gives the user the option to use less filtering for faster response.

We can argue about performance though. Later ...

DaMichel:
mean shift code:  1162  usec
basic com extraction:  801  usec
I think I can get rid of `cv::distanceTransform`, though. Masking areas using `frame_blobs` is almost as good.

Also I am an idiot as there is a little bug in this code ...

sthalik:
@DaMichel PT tracker at 75 Hz runs with 2% of CPU when the window isn't visible, and the CPU frequency is 3.06, rather than 3.5+ Turbo frequency.

Most of the cycles come from the webcam's DShow filter code.

At the very least allocate the float matrices in the `extract_points` function prologue. The extractor used to allocate everything each frame and it was way slower. I don't remember the numbers but VTune was clear on that.

DaMichel:
Oh well, I just scratched this part. It is not essential and to make it work, I'd have to make too many modifications to your code with doubtful benefit. 

I still like the mean-shift iteration however. I changed the stopping criterion so that less iterations are taken. In total the addition needs about 1/3 of the time the original.

I know you are busy but I would appreciate if you tried this patch eventually - without necessarily merging it.

sthalik:
I'll give the code more than shallow reading later. Now I need to leave soon.

DaMichel:
> Check for inequality. Camera resolution change is only done infrequently and this makes code more complex (see your roi usage). Reallocate like in extract_points, use as-is.

I wanted to be super smart and allocate only the space required to hold the current ROI. However, since this can change a little bit from frame to frame, I took the maximum. You suggest to allocate buffers as large as the full image?

> No need to warn about it. It's predictable that happens on resolution change and the original camera open.
Sure. I will remove all debug output. Also I will rebase everything into one commit.

> Use an epsilon here since 1) flush to zero 2) division by very small amounts is unstable.
Okay.

> I'm curious, what typical delta you get?
No idea. My experimental prototype showed the position to converge. So the delta becomes smaller and smaller.
I think an epsilon of 0.01 is reasonable. This is because, it corresponds to 0.1 of the size of a pixel, which is less than the average deviation from the known blob center in my synthetic benchmark.

> auto and -> ret-type is used when the ret type depends on decltype of the input parameters. You can use double and no -> ret-type.
Okay.

> Also use [&], not [=] since you don't mutate the captured lexical environment content.
Okay. But would it not be better style to capture `current_center` by value to avoid the indirection by the reference. I say style because the compiler is likely to do the right thing either way. I hope ...

> t_.start();
Already removed the timing code.

sthalik:
Practical? It doesn't copy by value from push lexical environment. Though copying a two-element struct isn't measurable. 

Add a checkbox, sure.

DaMichel:
> Practical? It doesn't copy by value from outer lexical environment. Though copying a two-element struct isn't measurable. 

I'd rather eliminate the indirection by the reference with a copy. I'm curious if there is a performance difference due to the reference versus a copy. However, I leave this for another time.

I'm going to add a checkbox and rebase.

DaMichel:
Done.

sthalik:
@DaMichel sorry but my SSDK broke and need to get the dev env back in shape. Fortunately I got a spare hdd and the other hdd with actual data wasn't broken.

sthalik:
What the hell anyway. Windows Crashing Experience so reboot, bios waits detecting AHCI - no SSD. Use a different PSU cable, a different connector, no difference.

DaMichel:
No problem. Good luck with the fixes.

sthalik:
@DaMichel we'll do it after 2.3.0 is released ok?

DaMichel:
I think this would be appropriate +1:

sthalik:
@DaMichel please add copyright.

sthalik:
I merged it manually in time for 2.3.0. It's an improvement for quality LEDs that shine evenly. It's a non-improvement for LEDs that shine brighter at their centers.

cf. 5196a67e23b24612b05d535cc8cf6797268c391c, b6d5f1708e7a5f27d2ac46a817ac5a45c1f692c3, and aa6b0e3841991272668ea02b1a7166c03e59e84a.

sthalik:
Note that for uneven LEDs it doesn't regress anything.

PR #442, opened 2016-09-05
Build issue on MSVC

DaMichel:
I get this error
```
CMake Error at cmake/opentrack-boilerplate.cmake:66 (message):
  no vcvars.bat run; probably no editbin in PATH
Call Stack (most recent call first):
  cmake/opentrack-boilerplate.cmake:85 (opentrack_fixup_subsystem)
  cmake/opentrack-version.cmake:48 (opentrack_compat)
  CMakeLists.txt:14 (include)
```
however the build works fine if I apply this commit. I think MSVC++ just sets the required variables for its child processes. Same for post/pre-build scripts. Library paths and so on are not system wide here.

Caveat: I didn't try to build all submodules of opentrack.


sthalik:
You're right, it only has a place in the toolchain file.

PR #443, opened 2016-09-05
filter-kalman parameter tuning

DaMichel:
Just a minor change: More deadzone when KF is unsure of the model. Still gives very smooth motion when head pose changes at near constant velocity.


sthalik:
👍

PR #448, opened 2016-09-08
tracklogging: begone file selector gui

DaMichel:
I didn't like it, considering it is mostly intended as a developer aid.
With this patch, the enable button is still there, but the edit field and the button are gone.
The file select dialog will pop up when the tracker starts.
The code is simpler, too.


sthalik:
If you prefer it that way then sure.

PR #453, opened 2016-09-15
Fix small compilation issue

MaximeMorel:
Small inconsistency here, the prototype is now:
static void options::detail::bundler::refresh_all_bundles()
So I remove the argument. Compilation fixed.


sthalik:
Thanks, seems like all stuff wasn't committed yet.

PR #496, opened 2016-11-16
Initial Dutch translations

MathijsG:
Initial Dutch translations


sthalik:
@MathijsG use the translation stub archive for more strings. But some were added/changed so it won't be complete without setting up a complete build environment.

sthalik:
@MathijsG it should be easy to set up a build environment now. Use Visual Studio 2015 Community, and generate a solution for it using cmake. Use prebuilt Qt.

PR #573, opened 2017-03-16
New possibilities for VR use same as non VR

wormeaten:
Support for multiple cameras support for positional tracking to encrease precision and angle view.
Also possibilities for multi source tracking. Similar as VR HMD so we could use arduino based tracking like EDtracker for directional tracking combined with other source true camera for positional tracking only.

This is an idea just to take in consideration. Any sort of freetrack based software need serious changes to stay on track with new hardware request. This could be achieved with already existing technology but with just different approach.


sthalik:
Please send an actual pull request and I'll review that.

Issue #352, opened 2016-05-09
support Valve's openvr for HTC Vive


PaulKeeble:
Would be really nice to get Vive support similar to the Oculus via SteamVR.

MyrddinE:
Second that. There are a lot of games that support TrackIR... it would be very pleasant to play them in my Vive with head tracking inside the Steam cinema, Virtual Desktop, or similar.

sthalik:
Can't we simply support all devices the SteamVR library supports?

sthalik:
openvr project link: https://github.com/ValveSoftware/openvr

We have here a "multiple CRT" problem. We can either live with that and avoid `std::` memory layout issues _or_ create a subprocess.

There's a version for OSX IA-32 so we should use a subprocess since our OSX build is amd64 only.

Bearnosen:
I´m eagerly awaiting this!! Any progress on this front?

Bobbehhh:
As am I! I just received my Vive and just assumed OpenTrack already had support for it. Can't wait till it does. Thanks!

sthalik:
Hey,

I'm working on support for an eye tracker right now. It's done except for relative view mode. Give it a few days and I'll finish it.

After I include SteamVR, someone needs to test it. It's also unlikely to work the first time.

MyrddinE:
I would be happy to test it. I have several games with full TrackIR support but no Vive support.

Bobbehhh:
I would also be happy to help test it. I have 16 years of experience in QA testing with video card HW/SW, game testing and general computer product development. Do you have a HTC Vive? Would having one help with getting this to work? if so, I'm sure we could setup some type of funding page (or something less public).

sthalik:
I appreciate the help guys. It's gonna be the next feature then.

Don't send me any money, please. We can talk on Skype once I have time, then you get the code.

The later Rift trackers with positional tracking I made just sending random improvements to the person testing it. It worked after few iterations.

Bearnosen:
Any progress? :)

IPFR33LY:
Lots of people who bought the Vive, having support for it within openTrack would be a huge bonus.

Bearnosen:
Any news? :smiley:

sthalik:
No, and you're not helping. Having me receive the notification each time an "are we there yet?" message is sent to github doesn't write any code.

Bearnosen:
I'm sorry, didn't mean to stress you out!

[...]

PR #597, opened 2017-04-09
tracker/steamvr: Fix rotation issues

antcodd:
Use well known matrix to euler decomposition formula, fixing some incorrect accesses. May need gimbal lock avoidance.
Don't use opentrack centering as ResetSeatedZeroPose preserves calibrated real world up (Y) vector.
Note: controllers are centered to HMD seated position.
fixes most remaining issues in #352

### DIFF
--- a/tracker-steamvr/steamvr.cpp
+++ b/tracker-steamvr/steamvr.cpp
@@ -250,19 +250,24 @@ bool steamvr::center()
     with_vr_lock([&](vr_t v, error_t)
     {
         if (v)
+            // Reset yaw and position
             v->ResetSeatedZeroPose();
     });
-    return false;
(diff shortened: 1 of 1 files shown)


antcodd:
References:
http://nghiaho.com/?page_id=846 which conveniently has a rotation matrix row major with the same axes definition as SteamVR.
http://www.songho.ca/opengl/gl_anglestoaxes.html
Similar to the code in [ftnoir_tracker](https://github.com/opentrack/opentrack/blob/unstable/tracker-pt/ftnoir_tracker_pt.cpp#L201)

antcodd:
ISC license is fine, not that it is a particularly large modification anyway.

sthalik:
Merging now. Let's deal with possible issues later.

Issue #577, opened 2017-03-20
linux hotkey description discrepancies


ljgdasfhk:
It is not possible to assign hot keys F1 to F12 and some others. They do not work when the window is minimized.
Only letters and numbers work well. But then the corresponding key does not work in other applications. For example, I can not write messages with this key.
In version 2.3, F1 F12 worked.

sthalik:
Using the X11 protocol, `XGrabKey(3)` doesn't allow for passing through the keys. Unless there's some way to inject the key back via Xlib or xcb without receiving the same key again, no can do.

Global shortcuts are typically done via desktop environment API but we have to be DE-agnostic here. Not to mention Wayland's gonna come and last time I checked there was no protocol for key sniffing.

We _could_ do it through evdev and bypass X11 silliness that way, but that makes for Linux-specific code rather than X11-specific code. Note that evdev needs device node permissions.

But _really_ I only support Linux to run Valgrind on the codebase. Linux is an OS for developers. It's not ready for end users. It probably won't be for several more years at least, or ever.

ljgdasfhk:
Используя протокол X11, XGrabKey (3) не позволяет передавать ключи. Если нет возможности каким-либо способом вставить ключ обратно через Xlib или xcb, не получив тот же ключ еще раз, это не удастся сделать.

Глобальные сокращения обычно выполняются через API среды рабочего стола, но здесь мы должны быть DE-агностиками. Не говоря уже о том, что Уэйланд придет, и в прошлый раз я проверил, что нет протокола для поиска ключей.

Мы могли бы сделать это с помощью evdev и обойти X11 глупости таким образом, но это делает код, специфичный для Linux, а не специфичный для X11 код. Обратите внимание, что evdev требует прав доступа к узлу устройства.

Но на самом деле я поддерживаю Linux только для запуска Valgrind в кодовой базе. Linux - это ОС для разработчиков. Он не готов для конечных пользователей. Вероятно, это произойдет не раньше, чем через несколько лет.

> Note that evdev needs device node permissions.
I run opentrack as administrator. It's not a problem.

> Linux is an OS for developers. It's not ready for end users. It probably won't be for several more years at least, or ever.

Many people already play on Linux. There is already a Steam client for Linux and games in it for end users.
I hope that you change your mind. I believe in your strength and experience. Good luck to you. I look forward to solving this small problem. The rest of the program works perfectly for the KDE.

sthalik:
[...]

PR #622, opened 2017-05-15
Linux/X11: Fix hotkeys

miniskipper:
See issue #577

### DIFF
--- a/qxt-mini/qxtglobalshortcut_x11.cpp
+++ b/qxt-mini/qxtglobalshortcut_x11.cpp
@@ -206,19 +206,44 @@ public:
         return DefaultRootWindow(display());
     }
 
-    bool grabKey(quint32 keycode, Window window)
+    bool grabKey(quint32 keycode, quint32 modifiers, Window window)
     {
         if (keybinding::incf(keycode))
         {
             QxtX11ErrorHandler errorHandler;
+            bool error = false;
 
-            XGrabKey(display(), keycode, AnyModifier, window, True,
+            XGrabKey(display(), keycode, modifiers, window, True,
                      GrabModeAsync, GrabModeAsync);
-
             if (errorHandler.error) {
+                error=true;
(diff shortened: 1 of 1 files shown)


sthalik:
@miniskipper please see if https://github.com/opentrack/opentrack/commit/f8634b81548833969b79f2f0673298dffcadba32 is to your liking. I implemented hashing keycodes together with modifiers. It also doesn't seem to be necessary to block the key with numlock and scroll lock.

If this is to your liking, I'll put my own changes on top of merging yours.

miniskipper:
@sthalik Sadly this does only work, if numlock on the keyboard is disabled. Thats why the key gets grabbed 4 times (num, scroll, both, none). Is it possible you are developing on a laptop? I wondered why hotkeys where working on my laptop - numlock is disabled there by default because of no dedicated numpad :)

sthalik:
On my PC104 keyboard, the "Num" modifier only gets raised if it's a number pad key being pressed. For other keys its state doesn't matter. I need to actually boot the Linux, something is amiss here.

sthalik:
@miniskipper we'd rather have a 1:1 mapping between keys with their modifiers, as recognized by Qt and X11. See how xev(1) reports KP_1 or KP_Home rather than the regular variants. It already exists for the Windows keyboard, see `logic/win32-shortcut.cpp`.

"Translating keysym to keycode" as it is right now is just a half-assed way that makes half the keys not work.

I don't have time to work on it soon but welcome contributions.

miniskipper:
The X Server treads the num lock key (and others) as a separate modifier. I think thats "normal"
For example have a look at the grabKey function in https://raw.githubusercontent.com/xfce-mirror/xfwm4/master/src/keyboard.c
I have seen code like this in almost every application that uses hotkeys under X.

sthalik:
:+1: :cat2:

PR #623, opened 2017-05-17
x-plane plugin

miniskipper:
macro rename
support head roll movement in plugin


sthalik:
Thank you

sthalik:
Do you wish to maintain Linux support in opentrack?

miniskipper:
Hi,
roll: when the tracking stops the roll stays at the last value. I found no easy possibility to control it with mouse/keyboard. Maybe it would be better to save the initial value instead of resetting to 0.

There is still some work to do:
- reinit view when plane is switched inflight (XPLM_MSG_PLANE_LOADED)
- dont move outside the plane
- add a possibility to save other views (passenger, copilot, ...)


Linux support: trying my best :)

sthalik:
Hey,

It's best you handle it if you want to do the changes.

Please make the change to allocate WineSHM as `static` as to avoid dynamic memory allocation via `calloc(3)`. The `.bss` section is always preinitialized to zeros.

Issue #667, opened 2017-08-03
bitrot of OSX-specific code


iDevPro:
1) (opentrack-compat/timer.cpp) Why you are use this overload in .cpp file compilation got error in there?
```
void Timer::otr_clock_gettime(timespec* ts)
**double Timer::otr_clock_gettime(timespec* ts)**
```
2) (opentrack-qt-mini/qxtglobalshortcut_mac.cpp) In function 
```
quint32 QxtGlobalShortcutPrivate::nativeKeycode(Qt::Key keys)
{
    UTF16Char ch;
    // Constants found in NSEvent.h from AppKit.framework
    switch (key) // would be "keys", because you pass keys as parameter, or rename parameter to key
    {
    case Qt::Key_Return:
        return kVK_Return;
    case Qt::Key_Enter:
...
```
Who can support OS X branch of code ?)

iDevPro:
Also in <cmath> abs want double, long double or float but not int:
opentrack-spline/spline.cpp
`x = std::abs(x); // x would be converted to double or long double ->  x = std::abs(double(x));`

iDevPro:
Also fix isNan isInf on mac:
append to nan.cpp
```
#elif defined(__APPLE__) && defined(__MACH__)
#   include <cmath>
#   define my_isnan std::isnan
#   define my_isinf std::isinf
extern "C" OTR_COMPAT_EXPORT bool __attribute__ ((noinline)) nanp(double x)
```

iDevPro:
But after build opentrack, install to folder, and "build from scratch" app. I can't use Face tracking and can't use plugin for x-plane :)

sthalik:
in re https://github.com/opentrack/opentrack/issues/667#issuecomment-320059066
```
Also function "abs" want double, long double or float but not int:
opentrack-spline/spline.cpp
x = std::abs(x); // x would be converted to double or long double -> x = std::abs(double(x));
```

No, see <<http://en.cppreference.com/w/cpp/numeric/math/abs>>.

in re https://github.com/opentrack/opentrack/issues/667#issue-247794756

> Why you are use this overload in .cpp file compilation got error in there?

No one runs OSX daily. It's supported worse than the Linux-specific code. Running it in a VM is a bitch too, even Mavericks.

We'd need the following at least to support as well as the Windows version:

- joystick input via QtGamepad. It's way easier than what Windows needs. Still, it hasn't been written.
[...]

PR #668, opened 2017-08-05
fixed macos build

MrCapone:
My specs: 
`macOS 10.12.6 Sierra` `Apple LLVM version 8.1.0 (clang-802.0.42)` `Cmake 3.9.0` `Xcode 8.3.3`

When i try build on macos using Clang i got some errors.

First on `opentrack/compat/timer.cpp:112:15`
```
error: return type of out-of-line definition of 'Timer::otr_clock_gettime' differs from that in the declaration
double Timer::otr_clock_gettime(timespec* ts)
```
_fixed by changing return type as declarated_.

Then i got linker errors for libopentrack-compat 
```
Undefined symbols for architecture x86_64:
  "isinf", referenced from:
      _nanp in nan.cpp.o
  "isnan", referenced from:
      _nanp in nan.cpp.o
ld: symbol(s) not found for architecture x86_64
```
_fixed by using  `isnan(double)` and `isinf(double)` from `<math.h>`_. This fix only for APPLE target.

Next error `opentrack/qxt-mini/qxtglobalshortcut_mac.cpp:174:14`
```
error: invalid use of member 'key' in static member function
```
_fixed by changing `quint32 nativeKeycode(Qt::Key keycode)` and `quint32 nativeModifiers(Qt::KeyboardModifiers modifiers)` to non-static functions._

Next `opentrack/spline/spline.cpp:148:9`
```
error: call to 'abs' is ambiguous
```
_fixed by using `<cstdlib>` instead `<cmath>`._

After those fixes i was able to build all project. Also i tested build on MS Visual Studio and Linux with gcc to prevert regression.

### DIFF
--- a/compat/nan.cpp
+++ b/compat/nan.cpp
@@ -19,6 +19,11 @@ int __cdecl my_fpclassify(double)__asm__("___fpclassify");
 #define my_isinf(x) (my_fpclassify(x) == FP_INFINITE)
 
 extern "C" OTR_COMPAT_EXPORT bool __attribute__ ((noinline)) nanp(double x)
+#elif defined __APPLE__
+#   include <math.h>
+#   define my_isnan(x) isnan(x)
+#   define my_isinf(x) isinf(x)
+extern "C" OTR_COMPAT_EXPORT bool __attribute__ ((noinline)) nanp(double x)
 #else
 int my_isnan(double)__asm__("isnan");
 int my_isinf(double)__asm__("isinf");
--- a/compat/timer.cpp
+++ b/compat/timer.cpp
@@ -109,7 +109,7 @@ mach_timebase_info_data_t Timer::otr_get_mach_frequency()
     return timebase_info;
 }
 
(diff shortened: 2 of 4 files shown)


sthalik:
```
error: invalid use of member 'key' in static member function
fixed by changing quint32 nativeKeycode(Qt::Key keycode) and quint32 nativeModifiers(Qt::KeyboardModifiers modifiers) to non-static functions.
```

This is incorrect. See <<https://github.com/opentrack/opentrack/issues/667#issue-247794756>>. The identifier name change `keys -> key` should be enough.

```
error: call to 'abs' is ambiguous
```

Good catch, thanks!

sthalik:
```
error: call to 'abs' is ambiguous
```

> fixed by using <cstdlib> instead <cmath>.

We need `<cmath>` for `std::fabs, fmin, fmax` et al. Good catch though.

MrCapone:
@sthalik 
Ok, i changed my commit (by amend) according to your remarks.

sthalik:
Thanks!

PR #673, opened 2017-08-08
 macosx/install-fail-tool: fixed install_name_tool targets path

MrCapone:
`install_name_tool` works only with dylib and binary files. No need work with subdirectories on `install-fail-tool` script.

### DIFF
--- a/macosx/install-fail-tool
+++ b/macosx/install-fail-tool
@@ -4,7 +4,7 @@ test -n "$1" || exit 1
 
 dir="$1"
 
-for i in "$dir"/* "$dir"/*/* "$dir"/*/*/*; do
+for i in "$dir"/*; do
         { test -x "$i" && test -f "$i"; } || continue
         case "$i" in
             *.dll|*.exe) continue ;;


MrCapone:
Example why i apply this little fix. Without it install-fail-tool go to sub directories (source-code doc etc) and get errors 

and with this fix everything goes smooth

sthalik:
It's not run on the Qt plugin in `platforms/` with this change, is it?

Try `otool -L filename` and see if any of your fully-qualified paths remain.

sthalik:
If the warning's the issue, please `chmod a-x` whatever gets picked up incorrectly. There's a check in `install-fail-tool`:

```
{ test -x "$i" && test -f "$i"; } || continue
```

so it's gonna help.

sthalik:
Try this instead:

[foo.patch.txt](https://github.com/opentrack/opentrack/files/1216776/foo.patch.txt)

MrCapone:
@sthalik I've been busy but now i check this patch. Yep, it works fine without any warnings.

sthalik:
@MrCapone It's merged into `unstable`. It's gonna work next time you build it.

PR #724, opened 2018-01-14
Fix XPluginEnable

PhilippMuenzel:
This fixes the plugin loading in an undefined state. See https://forums.x-plane.org/index.php?/forums/topic/138529-psa-check-your-int-xpluginenable-function-definition/ for more information


sthalik:
Thanks!

Issue #718, opened 2018-01-08
development of the Wii tracker


cpuwolf:
hi sthalik,
could you consider abstract Camera class that has the ability direct output 2D points, such as WII remote?
if so, you can put point::extract as part  job of class Camera. then we can  add WII remote supporting here.

actually I have duplicated tracker-pt code to support WII remote, it works good. but duplicate is not a good idea. hope you can consider supporting WII remote
  

![2018-01-11_175518](https://user-images.githubusercontent.com/1320329/34819581-0e23453c-f6f9-11e7-895e-c1d62d3d921e.jpg)

sthalik:
What do you want to replace in `tracker/pt` in total? I can make it into a template class, just make a list.

cpuwolf:
hi Sthalik,
I would like to instance a special for example WIImoteCamera from Class Camera. then I can almost reuse whole tracker_pt procedure.  so please abstract Class Camera has the ability direct output 2D points.

cpuwolf:
hi Sthalik,
as I look at code at 2.9.3, 'Class Camera' is 'struct Camera' now.

albertolg:
@cpuwolf , how do you connect the wiimote? bluetooth? isn't there an adapter or a DIY interface to connect with the usb?

sthalik:
I'm abstracting `Camera` and `PointExtractor` then. :+1:!

sthalik:
@cpuwolf the API is in `pt-api.hpp` and you can see how to make a module in `module.cpp`.

cpuwolf:
@sthalik I never think you are so fast👍. Thank you ver much. let me learn it

sthalik:
@cpuwolf please contact me here or via email if you have any questions. Don't worry, I'm a friendly person :smile:

cpuwolf:
@sthalik I inhert WIICamera and WIIPointExtractor in below link.

https://github.com/cpuwolf/opentrack/commits/pt-reuse

but I found WIImote doesn't work baseing on cv::Mat& frame. 
this special camera WIICamera should direct output std::vector<vec2>& points,  which I cannot do it in WIICamera.

then look like I have to make WIICamera and WIIPointExtractor they communicate with each other. sounds strange.

sthalik:
@cpuwolf,

Currently `cv::Mat` is used unconditionally for image preview. But it makes sense, we shouldn't require PT to use OpenCV's internal format for all camera types. I'm making the change.

[...]

PR #741, opened 2018-02-06
traker/wii: add wiimote auto pair with Windows Bluetooth Stack

cpuwolf:
#718 
I never thought this auto pair is so fast to be finished


sthalik:
❤️

PR #743, opened 2018-02-08
Update nl_NL.ts

MathijsG:
Updated some incorrect words and Russian pieces in the file.


sthalik:
Thanks!

@MathijsG you can also build opentrack up to the point of `nmake i18n`. Right now most strings are outdated as not to clutter the commit log. Our Russian translator runns `nmake i18n` on his own.

MathijsG:
Ah okay will dive into that later. Doesn't that require the whole stuff for all the compiling? Which will be a big bloat to my system. Can't we somehow autocompile those new translations on changes? :-)

sthalik:
@MathijsG I pushed the latest version as of now.

PR #746, opened 2018-02-09
tracker/wii: add lang chinese && BT optimization

cpuwolf:
I added chinese lang firstly. but it doesnt take effect
#718

### DIFF
--- a/tracker-wii/lang/zh_CN.ts
+++ b/tracker-wii/lang/zh_CN.ts
@@ -1,0 +1,293 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="zh_CN">
+<context>
+    <name>PT_metadata</name>
+    <message>
+        <location filename="../../tracker-pt/ftnoir_tracker_pt.h" line="+95"/>
+        <source>PointTracker 1.1</source>
+        <translation type="unfinished"></translation>
+    </message>
+</context>
+<context>
+    <name>UICPTClientControls</name>
+    <message>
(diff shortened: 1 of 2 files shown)


sthalik:
In toplevel `CMakeLists.txt`, there's

```
if(".${opentrack_all-translations}" STREQUAL ".")
    set(opentrack_all-translations "nl_NL;ru_RU;stub")
endif()
```
Add the language code after the semicolon.

If you have any problems ask, the issues are my fault really!!! I should fix it. The following is a description how to work around the brokenness for the time being.

- - -

Also see `cmake/opentrack-i18n.cmake`. You need to have the CMake cache variable `opentrack_disable-i18n-update` off. Finally, please try only run `nmake i18n-zh_CN`. If it doesn't work, run full `nmake i18n` and undo the changes to other translation files.

It's annoying like this because line numbers are stored in the i18n `.ts` files. If they were updated with each commit, that'd create an immense amount of line changes. Also if you changed the `.ts` for `ru_RU`, the translator for Russian would have problems.

sthalik:
After adding a new translation, your make tool will complain that `zh_CN.ts` files are missing. Copy `stub.ts` into `zh_CN.ts` for these nonexistent files. Use Cygwin or Powershell for that.

sthalik:
The old messages I sent don't apply anymore. It's fixed. Running `make install` will generate the translation automatically.

In particular, the `zh_CN.ts` files were generated only by adding to the language list in the toplevel `CMakeLists.txt`. There's no more work required anymore. `make i18n` runs automatically.

sthalik:
👍

Issue #748, opened 2018-02-12
opentrack now has lang RU and NL, can we add Chinese support?


cpuwolf:
I would love to support Chinese related portion. I think I could do it from module to module, and tracker/wii is just a beginning.

#746

sthalik:
Please see 6bb50c04e5c40a5e036e41dde11b1660c3b077f6. I fixed translation-related issues that we had.

I also added `zh_CN` to the language list in 31d9768413199f388c216d92d9db5158ce8785cf and then running `make install` created the `.ts` files automatically.

Finally, I fixed a bug where `make clean` deleted all translation files.

sthalik:
You should be able to do translations from now on. To update based on code changes, run one of the `i18n`, `all` or `install` targets.

It's best to run `QObject::tr` as long as you're within some `QObject` subclass.  See also `_(str, ...)` for when necessary to translate outside `QObject`.

Does it work now?

cpuwolf:
until latest commit: tracker/wii: silence #pragma deprecated

there is build error with target "INSTALL" in MSVC


1>------ Build started: Project: i18n, Configuration: Release Win32 ------
1>Building Custom Rule F:/works/GitHub/opentrack/CMakeLists.txt
1>CMake does not need to re-run because F:/works/GitHub/output/opentrack/CMakeFiles/generate.stamp is up-to-date.
1>Running lrelease for nl_NL
1>CUSTOMBUILD : lrelease error : Cannot open F:/works/GitHub/output/opentrack/variant/default/lang/nl_NL.ts: 系统找不到指定的路径。
1>Done building project "i18n.vcxproj" -- FAILED.

sthalik:
What happens if you run the `i18n` target manually?

This is strange because `/variant/default/lang/nl_NL.ts` exists in the repository.

sthalik:
I'm able to run:

```
dev/opentrack/build % find . -name '*.ts' -delete
dev/opentrack/build % msvc ninja install
```

While there are subtle differences between build systems, it should work after solution reload, `clean` or `mrproper`, then `install`. It should also work on following installs. Of course if you still have an issue it's unacceptable and I need to fix it.

cpuwolf:
I created a new folder, and fetched latest commit. now same error with target "build"

1>------ Build started: Project: i18n, Configuration: Release Win32 ------
1>Building Custom Rule F:/works/GitHub/opentrack/CMakeLists.txt
[...]

PR #751, opened 2018-02-21
add Chinese Lang

cpuwolf:
#748 
tracker/wii works
tracker/pt works
but main UI doesnt work


sthalik:
Add main to `variant/default`.

sthalik:
Option screens go to `gui` though.

sthalik:
Can you build with Ninja under "Visual Studio Command Prompt"? I'd like to ascertain whether it happens in all generators (.sln, ninja, makefile).

Also I heavily recommend using your own config as my own `sdk-paths-sthalik@MSVC-windows.cmake` in the root directory. opentrack has quite a few dependencies and that helps set them up immensely.

PR #772, opened 2018-04-28
wiiyourself worst problems

sthalik:
Hey,

The build emits a ton of warnings due to unbalanced parentheses in the original `IF_TYPE` definition. I used `__VA_ARGS__` to get around it. I also fixed some of the worst compile-time warnings.

I'd like to run `clang-format` on wiiyourself while preserving line authorship information. In the farther future, I'd like to add a wrapper similar to `std::shared_ptr` for `HANDLE` types, etc. so that they never leak with the user not needing to release them manually.

Please see if this set of changes broke anything. I have no wiimote to check.

### DIFF
--- a/tracker-wii/wiiyourself/wiimote.cpp
+++ b/tracker-wii/wiiyourself/wiimote.cpp
@@ -1773,50 +1773,54 @@ int wiimote::ParseReadAddress (BYTE* buff)
 				break;
 				}
 
-			#define IF_TYPE(id)	if(type == id) { \
-									/* sometimes it comes in more than once */ \
-									if(Internal.ExtensionType == wiimote_state::id)\
-										break; \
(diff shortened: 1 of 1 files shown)


cpuwolf:
OK. sthalik, I do find lots of problems in wiiyourself, let me check

sthalik:
@cpuwolf what do you think about RAII for Windows handles? Ideally I'd like to use partial specialization in order to avoid the need to specify the right destructor each time.

I don't know if the fallthrough with the warning is wrong but it probably is.

I'd like to machine-reformat the code (preserving `git blame`) before doing any further modifications. The code is badly formatted. opentrack is standardized on 4 spaces but this is all over the place.

cpuwolf:
it works

PR #869, opened 2019-02-03
Tracker kinect face

Slion:


sthalik:
@Slion welcome aboard! 👍

This API is strange. I can see it doing `malloc` a lot in the main loop by your code. Guess it can't be helped. I'm unable to test it at runtime due to lack of any Kinect hardware.

I found only minor style issues in your code. Let's merge this and deal with the issues in-tree.

PR #914, opened 2019-03-19
Kinect support for Point Tracker

Slion:
Kinect tracker now provides IR camera for Point Tracker to use.

### DIFF
--- a/tracker-kinect-face/camera_kinect_ir.cpp
+++ b/tracker-kinect-face/camera_kinect_ir.cpp
@@ -7,18 +7,14 @@
 
 #include "camera_kinect_ir.h"
 
-#if __has_include(<opencv2/core.hpp>)
+#ifdef OTR_HAVE_OPENCV
 
 //#include "frame.hpp"
 
 #include "compat/sleep.hpp"
-//#include "compat/camera-names.hpp"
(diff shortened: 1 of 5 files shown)


tracker-kinect-face/camera_kinect_ir.cpp
```diff
@@ -84,7 +80,7 @@ bool CameraKinectIr::is_open()
 void CameraKinectIr::WaitForFirstFrame()
 {
     bool new_frame = false;
-    int attempts = 100;
+    int attempts = 200; // Kinect cold start can take a while
```
sthalik: That's at least 10-20 seconds in either case. Following that, it takes at least 20 seconds to start/stop a tracker in the worst case, with the UI freezing.

sthalik: Alternately you could use `compat/timer` rather than measuring attempt count. Use a minimum attempt count and a minimum time period.

sthalik: The issue here is that the PT tracker won't terminate in its destructor, freezing the entire app. Ideally the program's logic could return false in `get_frame`, returning true in `start` unconditionally.

PR #922, opened 2019-03-27
Using standard _ITERATOR_DEBUG_LEVEL by default

### LINKED ISSUE #920: How to use OpenCV in Debug on Windows?
When using OpenCV DLLs on Windows I get a crash in Point Tracker cv::putText.
That's true for the official OpenCV distribution and when compiling it myself.
It seems the std::string parameter is not passed properly to the DLL, different calling convention?

Therefore I built OpenCV as a static library to use it thus in OpenTrack. Unfortunately I then hit a linker problem in OpenTrack which looks like this:
`Error	LNK2038	mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '2' doesn't match value '0' in camera.obj	opentrack-tracker-pt	C:\Dev\GitHub\bin\opentrack-64\tracker-pt\module\opencv_imgproc410d.lib(resize.obj)`

Both OpenCV and OpenTrack used CMake generator `Visual Studio 15 2017 Win64`.
Any clue what could be going on?
Any chance getting instructions on how to build OpenCV for debug?

Slion:
Thus making default MSVC build working with default OpenCV static library build.
Closes #920.

### DIFF
--- a/cmake/msvc.cmake
+++ b/cmake/msvc.cmake
@@ -20,6 +20,7 @@ SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 #C4928 - illegal copy-initialization, more than one user-defined conversion has been implicitly applied
 set(warns-disable 4530 4577 4789 4244 4702 4530 4244 4127 4458 4456 4251 4100 4702 4457)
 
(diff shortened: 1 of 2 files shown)


sthalik:
LGTM :+1:

PR #932, opened 2019-04-14
Easy Tracker v0.1

Slion:


sthalik:
@Slion how are we doing on the tabs vs spaces front?

sthalik:
As for the build failing, I'm gonna take care of it with some cmake-fu.

Slion:
> how are we doing on the tabs vs spaces front?

Well it should be all space really. Spotted any issue?

sthalik:
@Slion feel free to merge once you get rid of all excess duplicate code. Any other issues are secondary. Can you think of a better name though? :) How about "Light Tracker" or "LED tracker"? In any case I'd like having the name imply what kind of device is used for tracking. "Easy Tracker" is generic.

Slion:
> feel free to merge

Thanks 😄 

>  once you get rid of all excess duplicate code

Still mostly settings and UI. Not worth creating a common library over it as they will eventually go their on way anyway.

> Can you think of a better name though? :) How about "Light Tracker" or "LED tracker"? In any case I'd like having the name imply what kind of device is used for tracking. "Easy Tracker" is generic.

Actually I did not want to make it device specific as it could evolve really. Therefore I thought I rather use a "brand" name for it.

Slion:
It's starting to look really good. I'll be merging that soon.

sthalik:
@Slion good! Can you think of a good name for it?

Slion:
> Can you think of a good name for it?

Easy is as good a name as any.
See https://github.com/opentrack/opentrack/pull/932#issuecomment-484526281

sthalik:
> Actually I did not want to make it device specific as it could evolve really. Therefore I thought I rather use a "brand" name for it.

The issue here is that a light blob tracker isn't generic as-is. See the amount of non-light-tracking trackers here.

I'll leave it to your discretion.

Slion:
> The issue here is that a light blob tracker isn't generic as-is. See the amount of non-light-tracking trackers here.

I see what you mean.

> I'll leave it to your discretion.

Thanks, I'll leave it as it is for now. Re-branding could still be done later on if and when we judge it necessary.

sthalik:
I'm curious, does Kalman help you much?

Slion:
> I'm curious, does Kalman help you much?

I left it in, but I don't think it brings much if anything at all. Probably because I don't know how to use it.

Slion:
It could also be that OpenCV Kalman Filter implementation is primarily intended to be used to predict positions when loosing signal rather than do any filtering. Though is theory if you can detect bad or noisy poses then you could use the predicted one from the KF instead. Interesting thought, I guess that's how it ought to be used.

sthalik:
Note that Kalman prediction for filtering suffers from its own share of problems, e.g. overshoot, sluggish response. I'd advise against settings that trust the model more than the raw data. You're better off ignoring bad frames entirely.

PR #941, opened 2019-05-02
Kinect IR camera was not listed any more.

Slion:


tracker-kinect-face/camera_kinect_ir.cpp
```diff
@@ -36,12 +47,16 @@ namespace Kinect {
 
     std::vector<QString> CamerasProvider::camera_names() const
     {
-        IKinectSensor* ptr;
-        BOOLEAN b;
-        if (SUCCEEDED(GetDefaultKinectSensor(&ptr)) && SUCCEEDED(ptr->get_IsAvailable(&b)) && b)
+        IKinectSensor* kinect;
```
sthalik: Please fix it so that Kinect doesn't show with the SDK installed but when the device doesn't exist (like my machine).

Slion: That's the intention. Is it not working?

PR #946, opened 2019-05-05
Make sure Kinect IR sensor is not registered if no Kinect is found

Slion:


Slion:
Rebuilding `camera-names.cpp` is not ideal but that was the easiest way to implement that for me.
Ideally `camera-names.cpp` should live in its own library since it does not have any dependencies on OpenCV anyway.
With Kinect SDK there is no easy way to tell if Kinect is connected without opening it including potentially long delays which would be a pain to manage when registering cameras.

sthalik:
Thanks!

PR #994, opened 2019-10-21
proto/wine: support alternative wine installation.

jp7677:
This commit introduces three optional environment variables:
- WINE
- WINE_PATH
- WINE_LD_LIBRARY_PATH

These can be used run the wine wrapper with a non-system wine installation.
Most promenent use case would be using opentrack together with Proton.

PR #995, opened 2019-10-25
proto/wine: support alternative wine installation

jp7677:
This commit wraps the wine wrapper in a shell script.
That shell script can be modified to use a non-system wine installation.
Most prominent use case here is using opentrack together with
SteamPlay/Proton without the need to recompile opentrack.


jp7677:
This is based on the discussion from https://github.com/opentrack/opentrack/pull/994

sthalik:
How about I introduce a dropdown allowing for Proton support? The binary is always reachable through `~/.steam` so there's no need for presenting any further options.

sthalik:
See #996.

Issue #807, opened 2018-07-03
Just another noob question on FreeTrack and x64 games


HornetMaX:
Hi all,
I havew been using OpenTrack with an EdTracker in some bike sims (GP Bikes, MX Bikes) with excellent results via FreeTrack interface. Recently MX Bikes went x64 and now I'm stuck.

What are the options here ? get a freetrackclient64.dll somewhere ? Maybe using TrackIR interface (as it seems OpenTrack does provide the x64 dll for that) ?

sthalik:
I don't believe there's any 64-bit "freetrackclient" support. Even if opentrack provided that, nothing uses it.

NPClient64.dll is known to work across many titles. That use it, of course.

HornetMaX:
Well I guess apps used to freetrackclient.dll going to x64 would gladly have a freetrackclient64.dll instead of switching to something else :)
Anyway, I get that your suggestion is for the hosting x64 app to simply use NPClient64.dll instead. Thx !

jp7677:
> I don't believe there's any 64-bit "freetrackclient" support. Even if opentrack provided that, nothing uses it.
> 
> NPClient64.dll is known to work across many titles. That use it, of course.

Hi @sthalik 
I'm sorry for resurrecting this old issue. Unfortunately Elite Dangerous/Horizons x64 (via SteamPlay/Proton/wine on Linux) seems to want a freetrackclient64.dll, at least on my system . Looking at the wine logs the game loads `NPClient64.dll` but also tries to load `freetrackclient64.dll` which it cannot find. Head-tracking didn't worked on my system, but started to work immediately once I placed `freetrackclient64.dll` in the `opentrack/libexec` folder. I took the `freetrackclient64.dll` from the setup package provided at https://sourceforge.net/projects/facetracknoir/files/
Could you please consider including a `freetrackclient64.dll` like you did with the `freetrackclient.dll`?

sthalik:
@jp7677 I'll look into it.

PR #996, opened 2019-10-27
add support for proton wine

sthalik:


sthalik:
@jp7677 does this work for you ingame?

proto-wine/proton.cpp
```diff
@@ -0,0 +1,64 @@
...
+        ":HOME/.local/share/Steam/ubuntu12_32/steam-runtime/i386/usr/lib"
+        ":HOME/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/lib/x86_64-linux-gnu"
+        ":HOME/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/lib"
+        ":HOME/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/usr/lib/x86_64-linux-gnu"
+        ":HOME/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/usr/lib"
+    );
+    library_path += ':'; library_path += qgetenv("LD_LIBRARY_PATH");
+    ret.insert("LD_LIBRARY_PATH", library_path);
+
+    ret.insert("WINEPREFIX", expand("HOME/.local/share/Steam/steamapps/compatdata/805550/pfx"));
```
jp7677: The WINEPREFIX environment variable is different per game, proton create a separate prefix per game.
I guess this line should be removed here and a user would have to set this one before starting opentrack.

proto-wine/proton.cpp
```diff
@@ -0,0 +1,64 @@
...
+    ret.insert("WINEESYNC", "1");
```
jp7677: The WINEESYNC environment variable may differ per game, "1" is the default, but some games require ESYNC to be turned off. Within Steam a user can set this by e.g. the game launch options.
A user would need the possibility to override this one, either by UI, or by setting the same environment variable before starting opentrack. The same would apply to WINEFSYNC.
See https://github.com/ValveSoftware/Proton/blob/proton_4.11/proton#L518 and https://github.com/ValveSoftware/Proton/blob/proton_4.11/proton#L518

proto-wine/lang/stub.ts
```diff
@@ -4,19 +4,19 @@
...
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>OK</source>
+        <source>Wine (system)</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Cancel</source>
+        <source>Proton (Steam)</source>
```
jp7677: Officially it is `Steam Play`, see https://github.com/ValveSoftware/Proton/blob/proton_4.11/README.md#getting-started-with-proton-from-steam-play
May be "Proton (Steam Play)"?

sthalik:
Fair enough. Another issue I have is that using recent winelib results in incompatible wineserver protocol version.

Are you able to get it to work at all using this code, other than appid or other minor changes?

jp7677:
@sthalik I have just compiled and tried to get it running with Proton 4.11 and Assettio Corsa Competitione, but unfortunately it does not work. I see that the wrapper is started, but I have no headtracking in ACC.

sthalik:
I took your environment variable list from https://github.com/opentrack/opentrack/pull/994#issuecomment-544804390. Could it be that `WINE_LD_LIBRARY_PATH` has some missing entries?

jp7677:
Your branch works now here, but to make it work I had to set WINEPREFIX and WINEESYNC before starting opentrack. Weird.

jp7677:
To answer your question. I used this wrapper script for my second PR:
```bash
#!/bin/sh
PATH="/home/myuser/.local/share/Steam/steamapps/common/Proton 4.11/dist/bin/:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/bin:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/usr/bin:/usr/lib64/qt-3.3/bin:/usr/share/Modules/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin" \
WINEDLLPATH="/home/myuser/.local/share/Steam/steamapps/common/Proton 4.11/dist/lib64//wine:/home/myuser/.local/share/Steam/steamapps/common/Proton 4.11/dist/lib//wine" \
LD_LIBRARY_PATH="/home/myuser/.local/share/Steam/steamapps/common/Proton 4.11/dist/lib64/:/home/myuser/.local/share/Steam/steamapps/common/Proton 4.11/dist/lib/:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/pinned_libs_32:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/pinned_libs_64:/usr/lib64/atlas:/usr/lib64//bind9-export:/usr/lib64/llvm5.0/lib:/usr/lib64/iscsi:/usr/lib64/llvm6.0/lib:/usr/lib64/qt-3.3/lib:/usr/lib64/vtk:/usr/lib/wine:/usr/lib64/wine:/lib:/lib64:/lib/sse2:/lib64/sse2:/lib64/tls:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/i386/lib/i386-linux-gnu:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/i386/lib:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/i386/usr/lib/i386-linux-gnu:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/i386/usr/lib:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/lib/x86_64-linux-gnu:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/lib:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/usr/lib/x86_64-linux-gnu:/home/myuser/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/usr/lib" \
"/home/myuser/.local/share/Steam/steamapps/common/Proton 4.11/dist/bin/wine" ../libexec/opentrack/opentrack-wrapper-wine.exe.so
```
and i had started opentrack with
```bash
 #!/bin/bash
WINEESYNC=1 \
WINEPREFIX="/home/myuser/.local/share/Steam/steamapps/compatdata/805550/pfx/" \
/home/myuser/bin/opentrack/bin/opentrack
```
Considering I got your branch working here I guess that WINEDLLPATH is indeed not needed.

(Though I still don't understand why I have to set WINEESYNC and WINEPREFIX with your PR)

sthalik:
> (Though I still don't understand why I have to set WINEESYNC and WINEPREFIX with your PR)

That doesn't make much sense, yeah. Try launching `env > /tmp/foo` and see if they get properly set.

Does it make a difference whether you start opentrack before or after running the game?

jp7677:
Found it, you assembled the environment, but didn't yet applied it to the wrapper, thus a line like this is missing in your PR:
```
wrapper.setProcessEnvironment(env);
```

proto-wine/proton.cpp
```diff
@@ -0,0 +1,64 @@
...
+    QString path = expand(
+        "HOME/.local/share/Steam/steamapps/common/Proton PROTON/dist/bin:"
+        "HOME/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/bin:"
+        "HOME/.local/share/Steam/ubuntu12_32/steam-runtime/amd64/usr/bin"
+    );
+    path += ':'; path += qgetenv("PATH");
+    ret.insert("PATH", path);
+
+    QString library_path = expand(
+        ":HOME/.local/share/Steam/steamapps/common/Proton PROTON/dist/lib"
```
jp7677: Some consistency nitpicking: The colons here are at the beginning of each line whereas the colons are at the end for the path string elements.

jp7677:
FYI, with your latest commit I can use opentrack together with Proton and Assetto Corsa Competione without setting any environment variable before starting opentrack.

jp7677:
As I said above, Proton itself also sets WINEDLLPATH. I don't know what is wise here.
- Setting WINEDLLPATH for the opentrack wrapper too to ensure that the environment really looks the same. Or
- Leaving it out because (at least on my machine) it doesn't seem to be needed.

sthalik:
Bad news, looks like it's [necessary](https://github.com/zfigura/wine/search?q=do_esync&unscoped_q=do_esync) to enable esync/fsync on all processes running under the specific wineserver. This means more options to configure.

Another issue is the need to compile the Wine wrapper using the same wineserver version as the one used by Proton. This means installing either standalone Proton or Wine with that exact version, for the purpose of compiling opentrack's wrapper via winegcc.

sthalik:
I'm going to leave WINEDLLPATH undefined. It comes from regular Wine for specifying *additional* library paths, and Wine is capable of finding its own install prefix.

It should now work for arbitrary applications ids with esync/fsync configurable.

jp7677:
> Bad news, looks like it's [necessary](https://github.com/zfigura/wine/search?q=do_esync&unscoped_q=do_esync) to enable esync/fsync on all processes running under the specific wineserver. This means more options to configure.

Yes, this is what I meant with one of my first comments.

> Another issue is the need to compile the Wine wrapper using the same wineserver version as the one used by Proton. This means installing either standalone Proton or Wine with that exact version, for the purpose of compiling opentrack's wrapper via winegcc.

I'm on Fedora 30 here. My system wine is 4.17 (Staging). The opentracker wrapper compiled with that version works fine here with Proton 4.11. That said I understand what you mean. I guess the only good solution here would  be to make the folks at Valve aware of your project and ask them to include the opentrack wrapper in Proton itself.

> It should now work for arbitrary applications ids with esync/fsync configurable.

Yes, I have tested with both ACC and Elite Dangerous by only changing the Steam App ID. Both worked perfectly! (to be honest, for ED I had to copy the freetrackclient64.dll into the libexec folder first, but that is a different story).

On last thought, you might consider renaming "Application id" to "Steam App ID" or "Steam Application ID". I think one of these two term are more familiar to users. See https://developer.valvesoftware.com/wiki/Steam_Application_IDs

Thanks a lot for your work here, this is really looking very cool!

sthalik:
> for ED I had to copy the freetrackclient64.dll into the libexec folder first, but that is a different story

I better distribute the 64-bit version of freetrackclient then.

sthalik:
Can you see whether `freetrackclient64.dll` I sent is working?

jp7677:
> Can you see whether `freetrackclient64.dll` I sent is working?

Yes, it works with ED. Thanks and really cool. This also resolves https://github.com/opentrack/opentrack/issues/807

jp7677:
As a side note, not related to this PR:
I'm seeing that you put the opentrack configuration folder directly into the user home directory (`QStandardPaths::DocumentsLocation`). You could consider putting them into `~/.config/`, thus `QStandardPaths::ConfigLocation`, at least on Linux. This would fit somewhat better into the Linux conventions. I realize that this would be a breaking change.

sthalik:
For now let's allow `~/Documents/opentrack-foo` if it already exists.

Thank you for helping me along with Proton specifics.

jp7677:
It works perfectly now, thanks again for all your work here!

PR #1018, opened 2019-12-30
FreePIE Use packed struct with clang to receive datagram

rvt:
I noticed that in clang the datagram packed was expected to be 52byte. However the structure that is actually going over the wire is 50 byte.

This MR makes sure that on clang we also use the packed data structure


sthalik:
Make that:

```c++
#ifndef __clang__
#pragma pack(push)
struct
#else
struct __attribute__((packed))
#endif
{
...
}
#ifndef __clang__
#pragma pack(pop)
#endif

sthalik:
Thanks!

PR #1022, opened 2020-01-09
EDTracker support added.

ballista-milsim:
Based on Joystick tracker this module adds:
 * EDTracker calibration trigger on tracking start
 * EDTracker center trigger on Opentrack "center" tracker API event
 * Absolute axis mapping for pseudo 6dof (for example, roll left and right  can be mapped to always zoom-in left and right MFDs in DCS A10C)

### DIFF
--- a/tracker-edtracker/CMakeLists.txt
+++ b/tracker-edtracker/CMakeLists.txt
@@ -1,0 +1,6 @@
+if(WIN32 AND Qt5SerialPort_FOUND)
+    otr_module(tracker-edtracker)
+    target_link_libraries(${self} opentrack-dinput ${Qt5SerialPort_LIBRARIES})
+    target_include_directories(${self} SYSTEM PUBLIC ${Qt5SerialPort_INCLUDE_DIRS})
+endif()
+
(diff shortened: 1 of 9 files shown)


sthalik:
EDTracker already advertises as a joystick through HID. What's the benefit for communicating via serial?

ballista-milsim:
It's just and add-on with couple of features I find useful:
* Ability to send "commands" to serial port. In case of EDTracker they're center and reset tracking. 
* Ability to force abs() axis output. My use case is pseudo 6dof for left and right A10C MFDs:
** I mapped EDTracker (joystick) roll axis to x, y, z; When I roll right I want x move right, y move down, and z zoom in. When I roll left I want x move left, y move down, z zoom in. Therefore I need y and z always positive value and I haven't found how to do it with out of the box functionality.

Maybe abs() feature can be added to Options->Output and serial port commands to Options->Shortcuts or new configurable section. Please advise.

sthalik:
There's not enough rationale to merge it, as I'll be the one maintaining the code long term anyway.

erlking:
do you know if this patch should also work with edtracker pro?
not being able to center the device can be a real issue actually. (it does that when it gets plugged in/powered on, but games does not like that in runtime..)

Issue #968, opened 2019-07-05
tracker/easy & kinect wait for frame


sthalik:
@SLion,
There's an issue in tracker/easy where you specifically wait for 1000/Hz milliseconds for the next frame to arrive. Does this imply that the Kinect source returns the old frame rather than block until the next one arrives? That behavior wouldn't be in line with opentrack's video input API.

There are some issues where `iTicker` from tracker/easy complains about being destroyed from the wrong thread. Can you check this on your end?

Ideally you'd subclass `QThread` and run that kind of blocking code from overridden `QThread::run()`. This timer usage smells dubious.

I'd also like to get rid of iostreams but that's for another day.

Slion:
> There's an issue in tracker/easy where you specifically wait for 1000/Hz milliseconds for the next frame to arrive.

Where is that in the code?

> Does this imply that the Kinect source returns the old frame rather than block until the next one arrives?

I guess I'm just asking the Kinect API to provide me the current frame. I'm not sure what that API does but I doubt it blocks and waits for the next frame.

>  Can you check this on your end?

I guess I could whenever I get a chance.

> Ideally you'd subclass QThread and run that kind of blocking code from overridden QThread::run(). This timer usage smells dubious.

It was using `QThread` and spinning much like tracker-pt. Spinning was first quick fixed 1fb6122dce1a5bc117e925006d87828a52cfecb6 then properly addressed using a timer rather than a raw thread 088344f1bde14a24f90852316d01e6f1826da034.

> I'd also like to get rid of iostreams but that's for another day.

Ok, and another issue please otherwise it gets confusing 😃

sthalik:
>> There's an issue in tracker/easy where you specifically wait for 1000/Hz milliseconds for the next frame to arrive.

> Where is that in the code?

https://github.com/opentrack/opentrack/blob/unstable/tracker-easy/tracker-easy.cpp#L696

> I guess I'm just asking the Kinect API to provide me the current frame. I'm not sure what that API does but I doubt it blocks and waits for the next frame.

Please fix this. I ran into some deadlock with tracker/easy and can't fix this without having the Kinect camera provider work correctly.

> It was using QThread and spinning much like tracker-pt. Spinning was first quick fixed 1fb6122 then properly addressed using a timer rather than a raw thread 088344f.

With other camera input modules, tracker/pt spends most time sleeping.

Slion:
[...]

PR #1026, opened 2020-01-13
tracker/kinect: sleep when identical frame received

sthalik:
Implements #968


sthalik:
@Slion can you runtime test this on your hardware please?

Slion:
@sthalik Could be a while until I get around it. Would be faster if there was a binary I could download.

Slion:
Please go ahead, I'll just complain next time I get to use it if something is broken :)

PR #1037, opened 2020-01-23
WIP Tobii Eye tracker support.

ballista-milsim:
Please review and suggest how to improve.

### DIFF
--- a/tracker-tobii/CMakeLists.txt
+++ b/tracker-tobii/CMakeLists.txt
@@ -1,5 +1,6 @@
 # https://developer.tobii.com/download-packages/stream-engine-4-1-0-for-windows-x86
 # https://developer.tobii.com/download-packages/stream-engine-4-1-0-for-windows-x64
+set(SDK_TOBII "" CACHE PATH "Tobii SDK path")
 if(SDK_TOBII)
     otr_module(tracker-tobii)
     target_include_directories(${self} SYSTEM PRIVATE "${SDK_TOBII}/include")
--- a/tracker-tobii/CMakeLists.txt
+++ b/tracker-tobii/CMakeLists.txt
@@ -1,0 +1,16 @@
+# https://developer.tobii.com/download-packages/stream-engine-4-1-0-for-windows-x86
+# https://developer.tobii.com/download-packages/stream-engine-4-1-0-for-windows-x64
+if(SDK_TOBII)
+    otr_module(tracker-tobii)
+    target_include_directories(${self} SYSTEM PRIVATE "${SDK_TOBII}/include")
+
+    set(dll "${SDK_TOBII}/lib/tobii/tobii_stream_engine.dll")
+    set(lib "${SDK_TOBII}/lib/tobii/tobii_stream_engine.lib")
+
+    message(${self})
+    message(${dll})
+    message(${lib})
+
+    install(FILES ${dll} DESTINATION ${opentrack-hier-pfx})
+    target_link_libraries(${self} ${lib})
+endif()
--- a/tracker-tobii/ftnoir_tracker_tobii.cpp
+++ b/tracker-tobii/ftnoir_tracker_tobii.cpp
@@ -2,6 +2,8 @@
 #include "api/plugin-api.hpp"
 #include "compat/math.hpp"
 
+tobii::~tobii() = default;
+
 module_status tobii::start_tracker(QFrame*)
 {
     t.start();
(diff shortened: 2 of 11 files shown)


sthalik:
Don't use signals in the main loop. Queued signals have too much overhead. You can use a mutex.

Ignore the error signal since it's likely to be caused by intermittent device unplug.

Furthermore, the signals run inside the main loop since you don't establish any in the thread, which can't be accepted.

Initialize SDK_TOBII as a cache variable, cache type PATH, default to empty string.

The 'retries' and 'interval' variables could be initialized as static constexpr.

sthalik:
In case of startup error you can return module_status error message in case where startup takes less than several seconds. Otherwise use compat/run-in-thread to display an error message and stop the thread.

There's no body for the 'tobii' destructor. You must use QThread::wait with no timeout in any case.

amadeus:
What's the status on Tobii Eye Tracker 5 support? I've been doing repo searches, and from what I can gather, the foundation for the support had been added, but it's officially not quite implemented yet?

amadeus:
> Is it possible to input using Tobii now?

Only via this workaround, afaik: https://gist.github.com/amadeus/09b32e4d45f815f6d46686d367c4b6ee

sthalik:
Need somebody with the hardware to implement it. I previously tried it on an obsolete Tobii device but it wouldn't perform the same kind of tracking as PT, Aruco or even face trackers. It needs somebody with more experience in eye tracking, or just an intuition for it.

eyemack:
It would be really great if you could implement the headtracking of Tobii ET5. Eye tracking is not necessary for flight sims, etc.  FacetrackNOIR can do it, but it's a real pain having to run that and the superior Opentrack to get the Tobii working.

sthalik:
@eyemack, this needs a lot of fine tuning after the original implementation. Tobii tends to have two tracking modes: snap and accumulative. Which means lots and lots of patience from people with the actual hardware.

minecrawler:
Hello @sthalik! **I own a Tobii Eye Tracker 5** and would love to transform this brick into a working device on Linux. It's the only thing holding back my Desktop from switching away from Windows 11.

I have never developed any kind of bio-tracking. My CPP skills are limited - though I have 10+ years of professional coding experience in other languages.

I can deliver protocol traces (via WireShark) and if you prepare the knobs or guide me, I can try them out and tune them.

The Tobii Experience Suite does not work 100% on Windows 11, however as far as I can tell, the driver/engine do their job. If necessary, I can also install a Windows 10 or Linux to try different things on a spare SSDs or VM.

CypherDragon:
Unfortunately I'm not a developer of any kind (I'm just building basic Python skills at the moment) but I know that Talon Voice has somehow implemented eye tracking with Tobii hardware. There's no setup or any extra hoops really, either - it's pretty much plug-and-play under Linux, minus the addition of a set of udev rules to allow Talon to access the sensor. Once Talon is running, it will calibrate and run the mouse just fine with Tobii.

In the eye_mouse.py file included in the Talon package, it seems to be calling a 'tobii' module where it's able to pull info to correctly move the mouse. Could this implementation be as simple as asking Tobii for their SDK to add the driver/module required?

Some links to look at:
https://github.com/TalonCommunity/Wiki/blob/main/tobii_5.md <-- setup guide for Tobii 5 w/ Talon
https://talonvoice.com/ <-- Talon voice homepage
https://developer.tobii.com/ <-- Tobii SDK info

khoanguyen-3fc:
[tobii-forwarder.cpp](https://github.com/opentrack/opentrack/files/12675165/tobii-forwarder.txt)

I made a tiny cpp program to forward head pose signal from `tobii_stream` to opentrack `tracker-udp`, work great with my Tobii 4C

image
image

sthalik:
So you're saying that it already works?

khoanguyen-3fc:
It just my solution to the current situation:
- Tobii Development License
- I haven't had the chance to set up the development environment for OpenTrack.
- There's a bit of a learning curve involved in getting my hands on this project's codebase.

If anyone have binary built for this PR, maybe I can help test it (with my 4C).

sthalik:
I've sent a form to Tobii asking for access to the SDK used by this PR. Should get a reply within a few days.

> There's a bit of a learning curve involved in getting my hands on this project's codebase.

Try this -- https://github.com/opentrack/opentrack/wiki/Hacking-opentrack

khoanguyen-3fc:
Is it ok if I ignore the current code and start from scratch? It seems like `eyeware_beam` could be a good starting point.

sthalik:
Sure.

PR #1073, opened 2020-04-14
Fix the jump of Yaw and Roll when crossing the border +/-180 degrees

GO63-samara:
When crossing the border on Yaw or Roll +/-180 degrees, the Octopus makes a full rotation of 360 degrees. This made it difficult to use an inertial tracker in a wireless VR.


GO63-samara:
Bug iilustration:
https://youtu.be/NlPE0yR5014

sthalik:
Merged! Thanks.

PR #1095, opened 2020-06-05
Add proper color keying to tracker-pt

a1k0n:
Instead of just selecting the red/green/blue channels, also subtract the
other channels. This allows for point tracking with just colored spots,
rather than IR LEDs -- green and magenta work particularly well for
this.

This keeps the existing "Red/Green/Blue only" options, but adds
"Red/Green/Blue/Cyan/Yellow/Magenta chroma key" options, which do a better
job of isolating those colors.

Annotation 2020-06-05 103422

### DIFF
--- a/spline/spline-widget.cpp
+++ b/spline/spline-widget.cpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 
 #include <QPainter>
+#include <QPainterPath>
 #include <QPixmap>
 #include <QString>
 #include <QToolTip>
--- a/tracker-pt/lang/nl_NL.ts
+++ b/tracker-pt/lang/nl_NL.ts
@@ -232,6 +232,30 @@ Don&apos;t roll or change position.</source>
         <source>Green only</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Red filter</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Green filter</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Blue filter</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Cyan filter</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Yellow filter</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Magenta filter</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>pt_impl::TrackerDialog_PT</name>
--- a/tracker-pt/lang/nl_NL.ts
+++ b/tracker-pt/lang/nl_NL.ts
@@ -233,27 +233,27 @@ Don&apos;t roll or change position.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Red filter</source>
+        <source>Red chroma key</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Green filter</source>
+        <source>Green chroma key</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Blue filter</source>
+        <source>Blue chroma key</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Cyan filter</source>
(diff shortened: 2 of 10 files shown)


spline/spline-widget.cpp
```diff
@@ -5,6 +5,7 @@
 #include <algorithm>
 
 #include <QPainter>
+#include <QPainterPath>
```
a1k0n, line 8: this was necessary for it to compile under Qt 5.15.0, FWIW

a1k0n:
With this change I'm able to reliably track with just some lime green foam stickers glued onto a 3d printed clip (in red for contrast, but it doesn't matter):
https://twitter.com/a1k0n/status/1268729072198139904

tracker-pt/module/point_extractor.cpp
```diff
@@ -104,15 +104,11 @@ void PointExtractor::ensure_buffers(const cv::Mat& frame)
...
+void PointExtractor::filter_single_channel(const cv::Mat& orig_frame, float r, float g, float b, cv::Mat1b& dest)
 {
     ensure_channel_buffers(orig_frame);
 
-    const int from_to[] = {
-        idx, 0,
-    };
-
-    cv::mixChannels(&orig_frame, 1, &dest, 1, from_to, 1);
+    cv::transform(orig_frame, dest, cv::Mat(cv::Matx13f(r, g, b)));
```
sthalik, line 111: Did you benchmark this? I'd prefer a separate function for color mixing.

tracker-pt/lang/nl_NL.ts
```diff
@@ -232,6 +232,30 @@ Don&apos;t roll or change position.</source>
         <source>Green only</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Red filter</source>
```
sthalik, line 236: The difference between extracting a single channel and bias against that color isn't clear to the end-user.

sthalik: Consider "Red bias" as naming.

a1k0n: Agreed, the distinction between "Red only" and "Red filter" or "Red bias" is going to be hard to describe. It isn't a bias against the color, BTW, it's a bias against the others. So the meaning of "green filter" here is to filter out all colors other than green. Which sounds kind of like "green only".

Maybe "red/green/blue/cyan/yellow/magenta chroma key"? that's the term of art, technically.

a1k0n: I'll change it to chroma key. That's the right name for this.

a1k0n:
I wasn't able to get very high resolution performance data using the profiler; the new method is definitely more CPU intensive than just extracting a channel, obviously, and the impact is measurable, but it's not huge. I'm happy with the code right now but I will attempt to write a more rigorous benchmark and look for optimizations (since it doesn't really need to do a float dot product per pixel for this).

sthalik:
> the new method is definitely more CPU intensive than just extracting a channel, obviously, and the impact is measurable, but it's not huge.

People use opentrack with old CPUs, e.g. to play old IL2 Sturmovik. See how release binaries don't require AVX. PT is already very CPU-intensive. Just add the old extraction method back, and adjust the switch to use it in the old cases.

a1k0n:
I put the old extraction method back already, yes. I force-pushed the code so the change might not have showed up on Github correctly. Please take another look.

I did some benchmarking using QueryPerformanceCounter plus some filtering. Here's a comparison in some uncalibrated units:

 - Average method: 1.1e7
 - Natural method: 1.0e6
 - Red/etc only: 1.0e5
 - Chroma key (cv::transform floats): 5.0e6
 - Chroma key (cv::transform int8_t): 5.0e6

So it's about 5x as expensive as natural/one channel, and half as expensive as the "average" method.

sthalik:
Merged. Thanks!

sthalik:
FWIW, it's 50x as expensive.

> Red/etc only: 1.0e5
1 x 10^5
> Chroma key (cv::transform floats): 5.0e6
5 x 10^6

a1k0n:
Oh, sorry, that was a typo. natural and red/green/blue only are on par, clocking in at around 1e6.

a1k0n:
Anyways, thanks for merging!

PR #1103, opened 2020-06-13
Proton directory discovery

russell:
Hey, this is a WIP still but i thought i should put it up for feedback, my C++ is terrible, and i think there is more that could be done, but I'm not sure exactly the best way to go about it.

The intent of this PR is to use search paths to find proton wine and the app directory, so that we can support steam installations in multiple locations and 3rd party proton installations.

My Debian installation of steam has it in ~/.steam, and I'm not the only one, seem the latest something must have changed in the steam codebase.  The current GE build documentation puts steam in this directory too https://github.com/GloriousEggroll/proton-ge-custom#installation  but i don't want to remove the other paths because i don't know how much control the packages have over where steam puts things, so it might change again, or it could still be at the old location on some users PCs.  So having more search paths seems like a more sustainable solution.

So after much messing about, this works, but it has flaws, they could probably be overlooked, but they are real problems if you make any mistake in entering values or if you have old files lying around it could break.  What i think is missing is some way to warn the user if any of the values can't be found.  For example if the appid is wrong, or the steam runtime can't be found.  But i don't know the best way to do it if you could point me to somewhere else where your using a message box or something that would be sufficient to warn the user of the error i would be grateful.


proto-wine/ftnoir_protocol_wine_dialog.cpp
```diff
@@ -1,19 +1,32 @@
...
     ui.setupUi(this);
 
-    for (const char* version : proton_versions)
-        ui.proton_version->addItem(version, QVariant{version});
-
-    tie_setting(s.proton_version, ui.proton_version);
+    for (const char* path : proton_paths) {
+        QDir dir(QDir::homePath() + path);
+        dir.setFilter(QDir::Dirs);
+        dir.setNameFilters(proton_filter);
```
sthalik, line 22: Just use

```c++
dir.setNameFilters({ "Proton*" });
```

proto-wine/ftnoir_protocol_wine.h
```diff
@@ -12,6 +12,7 @@ using namespace options;
 #include <QString>
 #include <QProcess>
 #include <QMutex>
+#include <QtCore/QVariant>
```
sthalik, line 15: Don't use QtCore prefix.

sthalik:
> What i think is missing is some way to warn the user if any of the values can't be found. For example if the appid is wrong, or the steam runtime can't be found. But i don't know the best way to do it if you could point me to somewhere else where your using a message box or something that would be sufficient to warn the user of the error i would be grateful.

See: <<https://github.com/opentrack/opentrack/blob/unstable/tracker-udp/ftnoir_tracker_udp.cpp#L71>>.

russell:
OK, i have implemented the error handling using an exception class.  I hope that it's reasonable way to do it.  I was trying to avoid changing the signature `make_steam_environ` and making it return a C style bool.

I have also sorted some of the imports to do a minor clean, and spotted an error, with a missing return https://github.com/opentrack/opentrack/pull/1103/commits/8444c609936606a361dd2681a6e2356e0d5ee81c#diff-ba756345a920975aefcc7df048c0e6f6R93

proto-wine/ftnoir_protocol_wine.cpp
```diff
@@ -74,9 +75,13 @@ module_status wine::initialize()
         QString proton_path(const QString& proton_path);
 
         wine_path = proton_path(s.proton_path().toString());
-        env = make_steam_environ(s.proton_path().toString(), s.proton_appid);
+        try {
```
sthalik, line 78: Exception handling is disallowed on project level.

sthalik: Just use something like

```c++
using t = std::tuple<QString, QString, bool>;
auto [path, error_string, success] = ...;
if (!success)
    ...;
```

proto-wine/ftnoir_protocol_wine.h
```diff
@@ -24,8 +25,8 @@ struct settings : opts
                 esync{b, "esync", true};
 
     value<int>     proton_appid{b, "proton-appid", 0};
-    value<QString> proton_version{b, "proton-version", {} },
-                   wineprefix{b, "wineprefix", "~/.wine"};
+    value<QVariant> proton_path{b, "proton-version", {} };
```
sthalik, line 28: Is there a reason this is a QVariant? Otherwise it should be a QString.

russell: I tried both, only using QVariant works.  I'm trying to access the data attached to the item in the combox but when using a QString i always get the label.

The combo box labels are set to strings like `Proton 3.16` and the data field is set to `/home/russell/.steam/steam/steamapps/common/Proton 3.16`

I've been using  `tie_setting(s.proton_path, ui.proton_version);` to connect the proton_path to the combo box in both cases.

When i make the proton_path a QString
```
s.proton_path() == "Proton 3.16"
```

when i make proton_path a QVariant
```
s.proton_path().toString() == "/home/russell/.steam/steam/steamapps/common/Proton 3.16"
```

proto-wine/ftnoir_protocol_wine.cpp
russell: Thank you, that was exactly enough information for me to efficiently re-write this.

proto-wine/ftnoir_protocol_wine.cpp
```diff
@@ -71,15 +70,16 @@ module_status wine::initialize()
...
 
         wine_path = proton_path(s.proton_path().toString());
-        try {
-            env = make_steam_environ(s.proton_path().toString(), s.proton_appid);
-        } catch(const ProtonException &e) {
-            return error(e.getMessage());
-        }
+        qDebug() << s.proton_path().toString();
+        auto [proton_env, error_string, success] = make_steam_environ(s.proton_path().toString(), s.proton_appid);
+        env = proton_env;
```
russell, line 78: i tried using 
```
auto [env, error_string, success] = 
```

but i think because env is already declared on line 66, it was being lost when we exited the if scope.

the best solution i could find was to give it a different name and assign it to `env`

sthalik: Either that or use `std::tie`.

proto-wine/ftnoir_protocol_wine.h
sthalik: Then the combobox's userdata stores more information than the user can select. The user-visible option is "Proton 3.16" but the persistent setting is bound to an absolute path.

But that's a PITA. We've already gone through many revisions with this PR. I'm going to merge it and work on it later.

sthalik:
Merged. Thanks!

PR #1107, opened 2020-06-19
Add Linux Joystick Support

russell:
Add support for Linux joystick devices, this allows things like the EDTracker to
work, and possibly other devices.

I have only tested this in the Opentrack GUI so far, but all movements works.  There are some limitations, like only supporting 8 axes like the windows API.

To give a bit of context, I'm using the sysfs interface to find the joysticks because that will give me access to the device id, which is being stored in the guid field of the settings struct (like the windows version)  This GUID is then used to find the joystick when starting the driver, so that each time you click start it will find the joystick instead of hard coding the joystick device which could change.

This could be improved by reading the joystick information that comes from the joystick initialisation, but otherwise hard coding the limit of axes might be good enough for now.

### DIFF
--- a/tracker-linux-joystick/CMakeLists.txt
+++ b/tracker-linux-joystick/CMakeLists.txt
@@ -1,0 +1,4 @@
+if(LINUX)
+    otr_module(tracker-linux-joystick)
+    target_link_libraries(opentrack-tracker-linux-joystick)
+endif()
--- a/tracker-linux-joystick/ftnoir_tracker_linux_joystick.cpp
+++ b/tracker-linux-joystick/ftnoir_tracker_linux_joystick.cpp
@@ -1,0 +1,87 @@
+/* Copyright (c) 2013 Stanislaw Halik <sthalik@misaki.pl>
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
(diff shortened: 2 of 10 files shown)


tracker-joystick/ftnoir_tracker_joystick.cpp
```diff
@@ -46,13 +65,36 @@ void joystick::data(double *data)
...
+    for (int i = 0; i < 6; i++)
+    {
+        axes[i] = axes_state[i];
+    }
+#endif
     if (ret)
     {
         for (int i = 0; i < 6; i++)
         {
-            int k = map[i] - 1;
```
russell, line 55: I think i might have fixed a bug here, 

at the top of this method the joysticks axes are defined as 

```
    int map[6] = {
        s.joy_1 - 1,
        s.joy_2 - 1,
        s.joy_3 - 1,
        s.joy_4 - 1,
        s.joy_5 - 1,
        s.joy_6 - 1,
    };
```
that combined with the `int k = map[i] -1` means we are effectively subtracting 2

So the dropdown looked like this once it's values were evaluated in this loop
* Disabled == -2
* Axis 1 == -1
* Axis 2 == 0
...

sthalik:
Make a separate module out of this since it's too many #ifdefs.

russell:
OK, I have split it out to a separate module :D

sthalik:
LGTM

PR #1108, opened 2020-06-21
Add a new Hamilton filter

GO63-samara:
Hamilton filter key features:
- Instead of square, round (spherical) floating dead zones and smoothing areas are applied. Due to this, the angular size of these zones does not change when the Pitch angle changes. Diagonally rotations is as easy as moving along the Yaw and Pitch axes.
- Rotations are not filtered by independent coordinates, but comprehensively, in 3D space. Rotations and movements are more natural. There are no view jumps at the borders of +/- 180 degrees.
- The possibility of increasing the smoothing of rotations when zooming (when the head is approaching the monitor, that is, when increasing the -Z coordinate) is introduced. This makes it possible to more accurately aim and monitor remote targets.

A full description of the Hamilton filter is available in Russian here:
https://sites.google.com/site/diyheadtracking/home/opentrack/opentrack-hamilton-filter

The Hamilton filter was tested by the Russian community, received positive reviews:
https://forum.il2sturmovik.ru/topic/5061-opentrack-актуальная-информация-по-проекту-решение-проблем-вопросы/page/24/
https://forums.eagle.ru/showthread.php?t=23280&page=249

### DIFF
--- a/filter-hamilton/CMakeLists.txt
+++ b/filter-hamilton/CMakeLists.txt
@@ -1,0 +1,1 @@
+otr_module(filter-hamilton)
--- a/filter-hamilton/OTнастройки.txt
+++ b/filter-hamilton/OTнастройки.txt
@@ -1,0 +1,7 @@
+  Max radius :  20,00	2,5
+   Smoothing :   2,07
+   Dead Zone :   0,00	0,2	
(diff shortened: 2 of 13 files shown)


sthalik:
LGTM.

PR #1127, opened 2020-08-15
Create GitHub Action workflow for building project on Windows, Linux and macOS

pbek:
This could serve as start for CI and later even release building and deploying for all platforms (not just Windows).
The workflow would be run on all branches and for all pull requests.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -1,4 +1,4 @@
-name: Build
+name: Build Release
 
 on:
   push:
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -47,18 +47,17 @@ jobs:
       # Build project
       #
 
-    - name: Run cmake
-      run: |
-        cmake CMakeLists.txt
-    - name: Build OpenTrack
-      run: |
-        cmake --build .
-
     - if: contains( matrix.os, 'macos')
-      name: Build macOS App
+      name: Build for macOS
       run: |
-        ${Qt5_Dir}/bin/macdeployqt $APP.app
-        git clone --depth=1 https://github.com/aurelien-rainone/macdeployqtfix.git
-        python macdeployqtfix/macdeployqtfix.py $APP.app/Contents/MacOS/$APP /usr/local
+        rm -rf build
+        mkdir build
+        cd build
+        cmake -DCMAKE_BUILD_TYPE=Release ..
+        make install
+        tree -dl install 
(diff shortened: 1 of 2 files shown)


pbek:
I lately also migrated my https://github.com/pbek/QOwnNotes away from AppVeyor and TravisCI for CI and release building. GitHub Actions are so much more convenient and faster. And you can serve all three platforms at the same time.

pbek:
For QOwnNotes I also build releases for all three platforms that result in a single GitHub release with a ZIP for Windows, an AppImage for Linux and a DMG for macOS.

See: https://github.com/pbek/QOwnNotes/blob/develop/.github/workflows/build-release.yml

pbek:
The workflow for this PR builds for all three platform, see https://github.com/pbek/opentrack/actions/runs/209854964

Slion:
That is what we need. Would be great if we could download the resulting Windows installer once the build is finished.

Cat-Lady:
Any progress on merging this, after... 4 years?

sthalik:
On OSX it's missing dependencies such as foohid. On Windows attempts at automated builds won't be enough until the move to Qt 6 when it can finally be included as a submodule. If you want to help with builds then try getting CI to pick up the non-public `depends-nonfree` repo in read-only mode.

d3x0r:
Would be nice to have a nightly windows build - I don't want to install Qt dev under windows - tryin to get xreal one pro tracking to work...

sthalik:
Agreed, would be good and in fact most of the work is already done. What isn't done:

- automatically install Visual C++ Insider Preview Build Tools on the CI machine
- figure out how to plug the nonfree-sdk repo into the build
- build qt during the build
- build opencv, aruco, libusb during the build
- build onnxruntime (will take ages)
- make an initial cmake cache for each of these dependencies (e.g. disable examples/tests build for opencv, disable most modules for qt)

Making incomplete or non-fully-optimized builds isn't worth it IMO. otherwise will just be unused and gather dust.

msupino:
+1 for this — the lack of macOS CI is a real friction point. I just opened #2173 (macOS camera-enumeration + OpenMP linking fix) and there's currently no automated way for reviewers to verify a macOS-touching PR actually builds, so it's all manual or "trust me." Having Actions cover Win/Linux/macOS would meaningfully unblock review of macOS portability work.

Happy to test the workflow against my branch (`msupino:macos-camera-enum-fix`) if it would help shake out macOS-specific build issues before merge.

PR #1130, opened 2020-08-16
(idea) do the app ignore new "signals" if theres 3 already

asotelles:
Hello, soo I like to smoke and fly at the same time(lol) and to do that I have to lock the opentack soo the infra red signal from the cigarette dont make the tracking crazy. Soo I was thinking that maybe could be on the code something like "ignore 4th new signal" or "if theres 3 signals on the camera, ignore new ones". Something like that. Thx for your good work


sthalik:
Pull request sent to wrong branch.

PR #1143, opened 2020-08-30
tracker/easy: Disable Kinect distortion coefficients usage

### LINKED ISSUE #1020: P3P solver in Kinect not accurate on Z-depth and also pitch and yaw
Hi, thanks for the amazing work you provided here,
I almost succeeded in using Kinect IR marker solution (Easy Tracker version) here,
with some little buggy stuff I can't solve.
The P3P solver in Kinect not accurate on Z-depth and also pitch and yaw,
I am guessing it's related to the cap/IR marker model?
I tried these 2 set up, one bigger and one smaller:
image
image

Both of them have a similar performance issue in Easy Tracker.
Thanks in advance!

### LINKED ISSUE #1141: Easy Tracker Vertices
Hello,

Very cool program. The basic Kinect Facetrack is awesome. I was so shocked at how much of a game changer head tracking is in DCS World. After looking into it a bit more, I decided to use Easy Tracker. I can't seem to get it working properly, though. I believe my main issue is being confused about the settings. First, when I click on the settings under Input: Easy Tracker 1.0, they are labeled as PointTracker Settings. Second, the model tab. I have three vertices selected, but I am confused about the measurements. 

Using the ones I've found in these forums before has not yielded good results, despite me using a hat very similar to the picture floating around here (A small ball wrapped in reflective tape on the top middle of the hat, and two more on the bill, one on each side, looks like Splinter Cell haha). Are you able to explain it a bit better? I am able to figure out the mapping, for the most part, but there is also a bit of roll input when I give yaw, and the X axis moves me in and out instead of side to side.

Any help would be appreciated. This is a sweet program and has been fun to tinker with over the last week, but I admit that I may need some help!

Cheers

Slion:
Thus notably fixing badly broken Z component in translation vector.

Close #1141
Close #1020

### DIFF
--- a/tracker-easy/tracker-easy.cpp
+++ b/tracker-easy/tracker-easy.cpp
@@ -154,8 +154,13 @@ namespace EasyTracker
         // 5 - Radial fourth order
         // 6 - Radial fifth order
         // 7 - Radial sixth order
-        for (unsigned k = 0; k < 8; k++)
-            iDistCoeffsMatrix(k) = (double)iCameraInfo.dist_c[k];
+        //
+        // SL: Using distortion coefficients in this way is breaking our face tracking output.
+        // Just disable them for now until we invest time and effort to work it out.
+        // For our face tracking use case not having proper distortion coefficients ain't a big deal anyway
+        // See issues #1141 and #1020
+        //for (unsigned k = 0; k < 8; k++)
+        //    iDistCoeffsMatrix(k) = (double)iCameraInfo.dist_c[k];
     }
 
 


sthalik:
Can you remove all data structures and code related to distortion coefficients? There are no other uses other than Kinect.

Slion:
> Can you remove all data structures and code related to distortion coefficients? There are no other uses other than Kinect.

Sure could but ideally we should put it back once we sorted out what is wrong with it. Removing everything would be counter productive in that respect.

PR #1173, opened 2020-09-19
Fix display of the Octopus pose in the Pose-widget.

GO63-samara:
In the extreme version of OpenTrack-2.3.12, the Octopus pose is still displayed incorrectly.
I have fixed pose-widget. A video of the corrected pose-widget is available here:
https://youtu.be/my4_VOwGmq4

Fixed bugs:
- The turns and movements of the Octopussy are now performed truly independently of each other, as it should be in 6DOF.
- When cornering, there is no "gimbal lock" at Pitch = +/- 90 degrees.
- Fixed directions of axes of rotations and positions. Now the Octopus pose displays the actual direction of view on the plane.
- Fixed display of the back (green) surface of the Octopus. Previously, it was displayed mirrored.

Additional features:
- Applied "perspective projection", the picture becomes more voluminous.
- Added lighting effect from above, with the same purpose.
- Added background fill for the widget. This makes it possible to see the borders of the widget.
- Added X and Y axes. This helps to estimate how far the Octopus is deviated from the center.
- Added [Mirror] checkbox, mirroring positions and rotations. It is often more convenient to observe the Octopussy's mirror pose.
- If before compilation in the file "pose-widget.hpp" include line 19: "#define TEST", then a rectangular frame will be drawn around the Octopus. This is useful when testing a pose-widget to assess distortion and size.

### DIFF
--- a/pose-widget/ReadMe.txt
+++ b/pose-widget/ReadMe.txt
@@ -1,0 +1,26 @@
+Hi everyone!
+
+In the extreme version of OpenTrack-2.3.12, the Octopus pose is still displayed incorrectly.
+The pose-widget is responsible for displaying the Octopus pose in OpenTrack. I have fixed this widget.
+
+Fixed bugs:
+- The turns and movements of the Octopussy are now performed truly independently of each other, as it should be in 6DOF.
+- When cornering, there is no "gimbal lock" at Pitch = +/- 90 degrees.
+- Fixed directions of axes of rotations and positions. Now the Octopus pose displays the actual direction of view on the plane.
+- Fixed display of the back (green) surface of the Octopus. Previously, it was displayed mirrored.
+
+Additional features:
+- Applied "perspective projection", the picture becomes more voluminous.
+- Added lighting effect from above, with the same purpose.
+- Added background fill for the widget. This makes it possible to see the borders of the widget.
+- Added X and Y axes. This helps to estimate how far the Octopus is deviated from the center.
+- Added [Mirror] checkbox, mirroring positions and rotations. It is often more convenient to observe the Octopussy's mirror pose.
+- If before compilation in the file "pose-widget.hpp" include line 19: "#define TEST", then a rectangular frame will be drawn around the Octopus. This is useful when testing a pose-widget to assess distortion and size.
+
+The corrected pose-widget now displays the Octopus pose correctly and can be used to check OpenTrack settings, sometimes even without launching the flight simulator.
+
+A video of the corrected pose-widget is available here:
+
+https://youtu.be/my4_VOwGmq4
+
+fixed by GO63-samara <go1@list.ru> <github.com/GO63-samara> 2020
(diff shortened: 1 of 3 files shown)


sthalik:
Hey!

Apologies for taking so long. I'm in no shape to be writing code right now.

Slion:
> Hey!
> 
> Apologies for taking so long. I'm in no shape to be writing code right now.

Get better soon ❤️

GO63-samara:
Was it COVID? I wish you health too!

sthalik:
No COVID. Is there anything critical I should take care of?

GO63-samara:
This is strange.
I checked it, and `pose_widget::resizeEvent()` is called only once for the entire duration of Opentrack. But `Pose_widget::paintEvent()` is called several times per second, even in the `Stop` state.
Does it really need to make the `QImage background` a member of the `pose_widget`?

sthalik:
> Does it really need to make the QImage background a member of the pose_widget?

The widget refreshes some 15 times a second during tracking. Allocations are expensive, and can lead to memory fragmentation. The general policy is to avoid allocations to help quad-core users and reduce fan noise.

GO63-samara:
> Pose widget refreshes some 15 times a second during tracking.

> The widget shouldn't refresh that often when tracking is stopped. There's no timer. Does the CPU usage register?

Now I have checked again:
`pose_widget::resizeEvent()` called only once for the entire duration of Opentrack operation.
`pose_widget::paintEvent()` called about 20 times per second only when **Start** is running.
And only the `Mirror checkbox` is updated all the time.
CPU load **0%** with input `UDP over network`.

> Allocations are expensive, and can lead to memory fragmentation. The general policy is to avoid allocations to help quad-core users and reduce fan noise.

It seems that we do not understand each other. Allocations of what? What do you have in mind?

If `QImage background` or `QPalette palette`, then it makes no sense to make them class members.
These are local variables, and after `setPalette(palette)` they are no longer needed.
I tried to transfer the `QImage background` and `QPalette palette` to the `struct OTR_POSE_WIDGET_EXPORT pose_widget final`, but that didn't change anything, and then they will take up memory until the pose_widget is destroyed, and that's worse.

Or do you mean something else?

GO63-samara:
New idea:

When a non-camera tracker starts, a novice user may be frightened by the **NO VIDEO** message. "**NO VIDEO** means something went wrong", he thinks.
I suggest for non-camera trackers to display one more pose-widget with raw data input instead of **NO VIDEO**.

How about this?

sthalik:
I should really get back to writing code. Somehow I imagined that the allocation happened in `paintEvent`, which is just not true. It was in `resizeEvent`. Sorry for the noise, and particularly, thanks for not accepting my pointless demand.

> When a non-camera tracker starts, a novice user may be frightened by the NO VIDEO message. "NO VIDEO means something went wrong", he thinks.

The message used to say "NO FEED" and it was confusing. The feedback was that the current message is an improvement. I'd leave it as-is unless there's actually a user who's confused.

PR #1268, opened 2021-05-05
Saner instructions for building ONNX Runtime for neuralnet tracker

DaWelter:
Looks like `cmake --install .\buildx64\Release` works after all.
It's not officially supported though, and it didn't work with `RelWithDebInfo`.
Still better than using NuGet?!

### DIFF
--- a/tracker-neuralnet/BUILD.md
+++ b/tracker-neuralnet/BUILD.md
@@ -9,26 +9,12 @@ Source location: https://github.com/microsoft/onnxruntime
 In order to build, execute `build.bat` as follows:
 
 ```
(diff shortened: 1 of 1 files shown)


sthalik:
You should now have repo access. I have nothing to add on the subject as I'm using the Ninja generator with MSVC.

DaWelter:
Hm, but before this tracker ever makes it into an official build, you'd have to build the onnx runtime, too.

Or maybe not. What would you do?

sthalik:
A compile-time dependency is perfectly normal. I'll manage.

DaWelter:
So I'm going to close this PR since it became outdated and the target branch is wrong.

PR #1305, opened 2021-07-31
Adding a choice of centering method

GO63-samara:
Opentrack now uses several centering methods.
Select the centering method from the drop-down list: `[Options] - [Shortcuts] - "Centering method"`.

- **Disabled**: Centering is disabled.
- **Point**: Point centering, the same as in OT-2.3.10, but without jumps when crossing the border +/- 180 degrees. Well suited for video trackers, although near the input |Pitch| = 90 it has strange behavior similar to "gimbal lock".
- **Wireless VR 360**: Centering using a centering matrix or quaternion, the same as in OT-2.3.11...OT-2.3.13. Ideal for wireless VR headset for 360 degree viewing, no "gimbal lock". But it requires precise installation of the sensor or clip on the head so that in the central position the Roll angle on the "Raw tracker data" panel is close to zero.
- **Roll compensated VR 360**: It is a development of the previous version. It does not require precise installation of the sensor on the head. The sensor or clip can be tilted along the Roll, even at a 45 degree angle. Unlike the first option, there is no "gimbal lock". This centering option is suitable in most cases, especially for owners of an IMU tracker and laptop, when the screen is installed below (or above) eye level.

Do not forget to click `[OK]` to save the centering method selection to the profile file.

Hint: Now it is possible to reset the centering without stopping the tracking. To do this, while tracking, switch to `[Disabled]`, then press `[Center]`. This will reset the centering for all centering options.

### DIFF
--- a/gui/lang/nl_NL.ts
+++ b/gui/lang/nl_NL.ts
@@ -360,6 +360,22 @@ Press &quot;clear calibration&quot; to remove any calibration data pertaining to
         <source>None</source>
         <translation type="unfinished">Geen</translation>
     </message>
+    <message>
+        <source>Centering method</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Point</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Wireless VR 360</source>
+        <translation type="unfinished"></translation>
+    </message>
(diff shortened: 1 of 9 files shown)


sthalik:
FWIW, the patch still applies cleanly after the pose offset fix.

PR #1306, opened 2021-08-01
Tracker neuralnet update onnxrt cmake (linux)

gdolle:
A proposal for minor changes to support system wide install though ONNXRuntime standard cmake module. (Tested on linux).
Set `ONNXRuntime_ROOT` for local build, i.e `cmake -DONNXRuntime_ROOT=/path/to/install`

NOTE: This PR provides a minimal working ONNXRuntime cmake module (request only `libonnxruntime.so`, no other components checked (providers, etc...). Official package may provide `ONNXRuntimeConfig.cmake` in the future.

See bellow to how to build and test the tracker with microsoft/ONNXRuntime (the cmake way)

### method1 - Build  from source (cmake)

```sh
# Version 1.8.1,  clang (->bug gcc) + system protobuf
git clone --single-branch --branch v1.8.1 git@github.com:microsoft/onnxruntime.git
cd onnxruntime
git submodule update --init --recursive
cmake -B workdir \
    -DCMAKE_BUILD_TYPE=Release 
    -DCMAKE_C_COMPILER=clang \
    -DCMAKE_CXX_COMPILER=clang++ \
    -DCMAKE_INSTALL_PREFIX=${HOME}/onnxruntime-linux-x64-1.8.1 \
    -Donnxruntime_ENABLE_PYTHON=OFF \
    -DONNX_CUSTOM_PROTOC_EXECUTABLE=/usr/bin/protoc \
    -Donnxruntime_PREFER_SYSTEM_LIB=ON \
    -Donnxruntime_USE_FULL_PROTOBUF=OFF \
    -Donnxruntime_BUILD_SHARED_LIB=ON \
    cmake
cmake --build workdir # -j4 (-jX X procs)
cmake --install workdir
# Tested on arch 
```

_Compilation may take some times_

### method2 - Use prebuild package

Get a [prebuild packages](https://github.com/microsoft/onnxruntime/releases) for your arch, for instance 

```sh
wget https://github.com/microsoft/onnxruntime/releases/download/v1.8.1/onnxruntime-linux-x64-1.8.1.tgz
tar -xzvf onnxruntime-linux-x64-1.8.1.tgz -C ${HOME}
```

## Build opentrack

To build opentrack with for a local ONNXRuntime install,

```sh
cmake -B work \
    -DCMAKE_BUILD_TYPE=Release \
    -DONNXRuntime_ROOT=${HOME}/onnxruntime-linux-x64-1.8.1 \
    -DSDK_WINE=ON \
    -DCMAKE_INSTALL_PREFIX=${HOME}/opentrack \
    .  
```

### DIFF
--- a/cmake/FindONNXRuntime.cmake
+++ b/cmake/FindONNXRuntime.cmake
@@ -1,0 +1,31 @@
+# FindONNXRuntime
+# ===============
+#
+# Find an ONNX Runtime installation.
+# ONNX Runtime is a cross-platform inference and training machine-learning
+# accelerator.
+#
+# Input variables
+# ---------------
+# 
+#   ONNXRuntime_ROOT            Set root installation.
+#
+# Output variable
+# ---------------
+# 
+#   ONNXRuntime_FOUND           Variable indicating that ONNXRuntime has been
+#                               found.
+#   ONNXRuntime_LIBRARIES       Library implementing ONNXRuntime
+#   ONNXRuntime_INCLUDE_DIRS    Headers for ONNXRuntime
+
+find_library(ORT_LIB onnxruntime
+    CMAKE_FIND_ROOT_PATH_BOTH)
+find_path(ORT_INCLUDE onnxruntime/core/session/onnxruntime_cxx_api.h
+    CMAKE_FIND_ROOT_PATH_BOTH)
+
+if(ORT_LIB AND ORT_INCLUDE)
+    set(ONNXRuntime_FOUND TRUE)
+    # For CMake output only
+    set(ONNXRuntime_LIBRARIES "${ORT_LIB}" CACHE STRING "ONNX Runtime libraries")
+    set(ONNXRuntime_INCLUDE_DIRS "${ORT_INCLUDE}" CACHE STRING "ONNX Runtime include path")
+endif()
--- a/cmake/FindONNXRuntime.cmake
+++ b/cmake/FindONNXRuntime.cmake
@@ -13,10 +13,9 @@
 # Output variable
 # ---------------
 # 
-#   ONNXRuntime_FOUND           Variable indicating that ONNXRuntime has been
-#                               found.
-#   ONNXRuntime_LIBRARIES       Library implementing ONNXRuntime
-#   ONNXRuntime_INCLUDE_DIRS    Headers for ONNXRuntime
+#   ONNXRuntime_FOUND           True if headers and requested libraries were found
+#   ONNXRuntime_LIBRARIES       Component libraries to be linked.
+#   ONNXRuntime_INCLUDE_DIRS    Include directories.
 
 find_library(ORT_LIB onnxruntime
     CMAKE_FIND_ROOT_PATH_BOTH)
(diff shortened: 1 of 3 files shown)


tracker-neuralnet/CMakeLists.txt
```diff
@@ -1,23 +1,19 @@
...
+find_package(ONNXRuntime REQUIRED)
 
+if(OpenCV_FOUND AND ONNXRuntime_FOUND AND OpenMP_FOUND)
     otr_module(tracker-neuralnet)
     target_include_directories(${self} SYSTEM PUBLIC 
-        ${OpenCV_INCLUDE_DIRS} "${ONNX_INCLUDE_DIR}")
-    target_link_libraries(${self} 
-        opentrack-cv "${SDK_ONNX_LIBPATH}" opencv_imgproc opencv_core 
+        ${OpenCV_INCLUDE_DIRS} "${ONNXRuntime_INCLUDE_DIRS}")
+    target_link_libraries(${self} "${ONNXRuntime_LIBRARIES}"
```
sthalik, line 10: A list of libraries must be used without double-quotation marks. It only works by accident because there's one element in the list.

sthalik (changes requested):

tracker-neuralnet/CMakeLists.txt
gdolle: @sthalik Just updated.

cmake/FindONNXRuntime.cmake
```diff
@@ -0,0 +1,40 @@
...
+# Output variable
+# ---------------
+# 
+#   ONNXRuntime_FOUND           True if headers and requested libraries were found
+#   ONNXRuntime_LIBRARIES       Component libraries to be linked.
+#   ONNXRuntime_INCLUDE_DIRS    Include directories.
+
+find_library(ORT_LIB onnxruntime
+    CMAKE_FIND_ROOT_PATH_BOTH)
+find_path(ORT_INCLUDE onnxruntime/core/session/onnxruntime_cxx_api.h
```
DaWelter, line 22: This path looks like a little issue because it does not match the directory structure of the official release package where I find everything is in one directory
grafik
Could you have both paths searched?

tracker-neuralnet/CMakeLists.txt
```diff
@@ -1,23 +1,25 @@
...
+        ${OpenCV_INCLUDE_DIRS} 
+        ${ONNXRuntime_INCLUDE_DIRS}
+        )
     target_link_libraries(${self} 
-        opentrack-cv "${SDK_ONNX_LIBPATH}" opencv_imgproc opencv_core 
-        opencv_imgcodecs opencv_calib3d
-        OpenMP::OpenMP_C)
-
+        opentrack-cv 
+        ${ONNXRuntime_LIBRARIES}
```
DaWelter, line 15: Since you defined a target for the onnx library, just use that instead (`onnxruntime `)

tracker-neuralnet/CMakeLists.txt
```diff
@@ -1,23 +1,25 @@
...
+find_package(ONNXRuntime QUIET)
 
+if(OpenCV_FOUND AND ONNXRuntime_FOUND AND OpenMP_FOUND)
     otr_module(tracker-neuralnet)
-    target_include_directories(${self} SYSTEM PUBLIC 
-        ${OpenCV_INCLUDE_DIRS} "${ONNX_INCLUDE_DIR}")
+    target_include_directories(${self} 
+        SYSTEM PRIVATE 
+        ${OpenCV_INCLUDE_DIRS} 
+        ${ONNXRuntime_INCLUDE_DIRS}
```
DaWelter, line 11: With  the target name in link_libraries this can be removed

tracker-neuralnet/ftnoir_tracker_neuralnet.h
```diff
@@ -26,7 +26,7 @@
 #include <memory>
 #include <cinttypes>
 
-#include <onnxruntime_cxx_api.h>
+#include <onnxruntime/core/session/onnxruntime_cxx_api.h>
```
DaWelter, line 29: Even worse issue with the path here. I guess it should not be too hard to add `<onnxlibrary include dir>` as well as `<onnxlibrary include dir>/onnxruntime/core/session/` to the include paths. Then this line need no change at all.

DaWelter:
I like it. I mean, why not. But it needs some work with the include paths.

gdolle:
@DaWelter thx for the comments. Yes I didn't notice they package differently from their cmake install target...  I'll fix that asap

P.S. Yes, I forgot to remove those variables (for other onnxrt usage ;) )

gdolle:
@DaWelter I update based on your comments. It works for me either using a released package or building from source (with ONNXRuntime_ROOT, or system wide install).
Could you check for windows from your side ? 
Don't hesitate if you see other problems.

PS. A github action pipeline would be nice in the future

DaWelter:
Hey @gdolle , it is working for me. Thank you. 

The only problem is that the onnxruntine dll is not installed. Looks like it never was. 

So @sthalik , I think we can merge this PR. Afterward I would see about the dll.

> PS. A github action pipeline would be nice in the future
Yeah, regarding that I refer to my fellow devs.

gdolle:
>  onnxruntine dll is not installed

@DaWelter Yes It was intended, I expect ONNXRuntime to be an external dependency. 
Feel free to add it, if you prefer to package onnxrt libs (dll and .so for win/osx/linux) within opentrack.

> Yeah, regarding that I refer to my fellow devs.

:+1: 

PS. I did not update the BUILD.md, we may have to update the doc somewhere.

PR #1307, opened 2021-08-03
cmake: improve the FindONNXRuntime script and install the ONNX dll

DaWelter:
@sthalik , here is the promised follow-up PR. There were a few things that should be corrected. I didn't notice before. Now the find script feels like overkill. Otoh it also finds the DLL which is nice. And the import target is nice, too. Suppose it is the right approach for modern CMake.

Please review.

Have to check the latest iteration on windows.

Here is an overview of the changes

* Add ONNXRuntime_DIR variable
* Change target name to onnxruntime::onnxruntime following naming
  conventions for imported libs
* Look for the DLL on windows and store it in a variable for later
* Correctly define the properties of the import library
* Change names and logic for global variables to follow standard conventions

@gdolle FYI

### DIFF
--- a/cmake/FindONNXRuntime.cmake
+++ b/cmake/FindONNXRuntime.cmake
@@ -8,7 +8,9 @@
 # Input variables
 # ---------------
 # 
-#   ONNXRuntime_ROOT            Set root installation.
+#   ONNXRuntime_ROOT            Set root installation. This is an environment
+#                               variable
+#   ONNXRuntime_DIR             CMake variable to set the installation root.
 #
 # Output variable
 # ---------------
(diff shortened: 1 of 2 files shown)


cmake/FindONNXRuntime.cmake
```diff
@@ -17,26 +19,63 @@
...
+        set_target_properties(onnxruntime::onnxruntime PROPERTIES
+            IMPORTED_LOCATION "${ONNXRuntime_RUNTIME}"
+            INTERFACE_INCLUDE_DIRECTORIES "${ONNXRuntime_INCLUDE_DIR}"
+            IMPORTED_IMPLIB "${ONNXRuntime_LIBRARY}")
     endif()
 endif()
 
-unset(ORT_LIB CACHE)
-unset(ORT_INCLUDE CACHE)
+mark_as_advanced(ONNXRuntime_INCLUDE_DIR ONNXRuntime_LIBRAY ONNXRuntime_RUNTIME)
```
sthalik, line 78: Typo in "_LIBRAY".

cmake/FindONNXRuntime.cmake
```diff
@@ -17,26 +19,63 @@
...
+
+set(ONNXRuntime_INCLUDE_DIRS ${ONNXRuntime_INCLUDE_DIR})
+set(ONNXRuntime_LIBRARIES ${ONNXRuntime_LIBRAY})
```
sthalik, line 81: Typo.

DaWelter:
Soooo ... now it seems to be working on both Linux and Windows.

The last push changes the library type from `UNKNOWN` to `SHARED` because without it cmake wanted to link against the dll where it should use the importlib.

Also fixed the typos of course.

DaWelter:
bump @sthalik. I'd say it's okayish now ;-)

PR #1319, opened 2021-08-28
Added translation for Centering method (RU)

GO63-samara:


sthalik:
Cool, thanks!

sthalik:
Do you feel like doing the rest of Russian strings before the release? We should be good to go.

PR #1327, opened 2021-09-04
Fix video capture on linux and mac as well as init of QCheckBox

DaWelter:
Noticed that video capture was broken after updating the OpenCV version. This PR brings back the `video_capture_backend` variable which is still defined, but has been unused for a while it seems. Now it's working again for me.

The other thing with the initialization looks harmless. The extra copy was not needed though. And my compiler complained about it.


sthalik:
Good catch, thanks!

Issue #1172, opened 2020-09-19
how to build for x-plane output


phiber2k4:

coljac:
This should be considered a bug, as the readme says that X-Plane plugin is an output mode, but as @phiber2k4 points out it does not show up in the output list.

After looking in the source, and following a hunch, I finally set the output to Wine and then it did work in X-Plane. I don't know quite why or what is happening here technically, but it did work. Can anyone confirm there's not supposed to be an X-Plane output?

sthalik:
It's using the Wine shared memory mapping for simplicity's sake. For that reason there's no duplicate output. Furthermore, you can build the Wine output without Wine itself, for that reason.

PR #1369, opened 2021-11-07
Adds a note regarding X-Plane plugin output (issue 1172)

coljac:
The readme states that Opentrack supports output to the X-Plane plugin. However, it is not documented anywhere that this is actually accomplished via the Wine output mode, leaving users [potentially confused](https://github.com/opentrack/opentrack/issues/1172) as to how to build Opentrack to enable an X-Plane output mode. I added a short note to this effect, which should steer would-be users in the right direction.

This resolves the above-mentioned issue 1172.


sthalik:
Thanks!

PR #1423, opened 2022-02-07
Adding input option to connect to IPhone FaceID Tracking via Unreal Engine's LiveLink App

Dazbii:
The iPhone Face ID sensors can accurately capture all sorts of facial expression data, including head rotation data. Unreal Engine has an app called Live Link Face that sends that is meant to send that data in real-time over to Unreal Engine. This data is just sent over UDP so it is easy to read and extract the head rotation data from.

I just added a new input, mostly copying the existing UDP input, but made it extract and format the data sent by the Live Link app.

One note is that Live Link only sends head rotation data, nothing on position.

Another thing to note is that the Face ID sensors are present on the iPhone X/XR/XS and newer, as well as the iPad Pro 3rd gen and newer.


sthalik (changes requested):

luzpaz:
Any reason why this PR is stalled ?

LCDraws:
any news on this?

Issue #1297, opened 2021-07-15
Max of the X Y Z 75


DimMario:
Max of the X Y Z 75.Is it possible to somehow expand from 75 to 200-300 . For use together with VR components?

GO63-samara:
Here's the same request: https://github.com/opentrack/opentrack/issues/1213

Issue #1213, opened 2021-01-04
[Feature request] Make the max. 75 CM of in-game movement or X Y and Z adjustable


MathijsG:
Hey Stanislaw,

I was wondering where the max in-game CM of 75 is based on? Could it be a possible feature to have make this adjustable? I see some games that could do more than 75 in-game CM's.

sthalik:
> I was wondering where the max in-game CM of 75 is based on? 

Back then there was no combobox allowing you to select the maximum output value. So adding more output made using the plot more inaccurate.

> Could it be a possible feature to have make this adjustable?

Yes, that makes sense. It's a 10-minute change, but I don't have the environment set up and it's going to be a while, honestly.

MathijsG:
Thanks Stanislaw,

Would be great if we could set these values in the future. Best would be to let it be user input?

For the physical Z-zoom axis the default selection options for the amounts of CM also doesn't always fit people's usecases either. Maybe we should make this configurable as well?

sthalik:
> Would be great if we could set these values in the future. Best would be to let it be user input?

I'm normally against things being *too* configurable. Anyway, what's the range of values you're interested in?

> Maybe I could change this myself, I'm unemployed next month :-P.

That's sad about your employment. Feel free to send a pull request.

I should release 2.3.13 but that needs the Wii tracker to be fixed first.

MathijsG:
I agree with things not always need to be too configurable, but this is one of the things where the use case can be so broad, and every game has a different amount of zoom, that's why I opt for every reasonable value via user input.

I don't know if there is a hard limit on the TrackIR protocol side of things? My use case with this feature was DCS, where I could just move 75 CM forward, while I could zoom in much further with another non-tracking axis. Maybe limit it to 999 CM max as user input?

sthalik:
> I don't know if there is a hard limit on the TrackIR protocol side of things?

No, it's just a value. Games clamp it though. Maybe install the TrackIR software? I'm not going to accept its EULA so can't do it myself.

MathijsG:
Checked it with my TrackIR which I bought for testing purposes. Doesn't look like Z is really limited. It goes to 300+ CM's. So a 3 digit value as max seems fine to me?

I've uploaded the sgl.dat to: https://mega.nz/file/DVQBTKLI#5iw3yB1Vpg8D496WmR-R9GUqQ26ifqrAfS7LjoTClm4

MathijsG:
Could this be implemented easily?

[...]

PR #1426, opened 2022-02-17
Added more max_clamp options

sharpTrick:
I added more "Max input" options as well as "Max output" options to the spline mapping GUI.
I specifically added  150 cm, 300 cm and 600 cm options.

I imagine that these options should more than enough to cover most use cases and solve the following feature requests:
#1213  
#1297

If these additional options are still not enough, by looking at my changes it should be easy for someone to figure out how to add 900 or even 1200 as additional options.


sthalik:
What's the use case here? Which game actually requires you 600+ cm of translation to move around the cockpit?

sharpTrick:
@sthalik, perhaps there is not one. The use case I imagined was one in which an individual could use opentrack as a relative position tracker from wherever their camera is set up. Potentially in room scale VR cave setups and for use cases beyond gaming.

I don't care what the official options are as I'm able to compile custom versions of opentrack for myself. 
Are there any changes you would like me to make to enable this merge for the purpose of solving the referenced requests?

sthalik:
It's a good idea, as no one will ask for increasing the limit ever again. Thanks.

JoDaEmPa:
Hello scharfTrick and stalik  ! I can't compile custom versions of Opentrack. It would be great if I could get the changes of 3 files. Or maybe it could be put in an extra version! with confidence the Robby ;-)

PR #1476, opened 2022-05-19
Neuralnet Tracker Improvements

DaWelter:
@sthalik , as promised.

It's a bit big so I left the individual commits instead of squashing.

There are some nice changes, I guess. :-)

Please test if possible because I changed how the face region is determined during tracking.

Been experimenting with recurrent neural nets, hence the support for those. But still I haven't found something that I like more than the released net.

### DIFF
--- a/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
+++ b/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
@@ -15,6 +15,7 @@
 #include <opencv2/calib3d.hpp>
 #include <opencv2/imgcodecs.hpp>
 #include "compat/timer.hpp"
+#include "compat/check-visible.hpp"
 #include <omp.h>
 
 #ifdef _MSC_VER
@@ -54,6 +55,18 @@ std::string convert(const QString &s) { return s.toStdString(); }
 #endif
 
 
+template<class F>
+struct OnScopeExit
+{
+    explicit OnScopeExit(F&& f) : f_{ f } {}
+    ~OnScopeExit() noexcept
+    {
+        f_();
+    }
+    F f_;
+};
+
+
 float sigmoid(float x)
 {
     return 1.f/(1.f + std::exp(-x));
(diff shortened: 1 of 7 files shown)


sthalik:
The changes do indeed look nice from reading the description and the code. I can't test it as my only visible light webcam has broken lens. We can merge it now and do a public test build while postoponing the next release, what do you think? The user complaining about MJPEG framerate should be willing to help.

From experience, the thing to watch out for is assertion failures when user's face is right next to the edge of the screen.

DaWelter:
> The changes do indeed look nice from reading the description and the code. I can't test it as my only visible light webcam has broken lens. We can merge it now and do a public test build while postoponing the next release, what do you think? The user complaining about MJPEG framerate should be willing to help.

Agreed.

Beware I put two more commits on top. Fixes - sort of.

> From experience, the thing to watch out for is assertion failures when user's face is right next to the edge of the screen.

This has always been fine. I think there is only one place where things could go wrong and `cv::getRectSubPix` takes care of the tricky parts.

tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -356,80 +587,138 @@ std::optional<PoseEstimator::Face> PoseEstimator::run(
...
+        // In old behaviour  ROI is taken from the model outputs
+        const vec3 offset = rotate_vec(face->rotation, vec3{0.f, 0.1f*face->size, face->size*0.3f});
+        const float halfsize = face->size/float(settings_.roi_zoom);
+        face->box = cv::Rect2f(
+            face->center.x + offset[0] - halfsize,
+            face->center.y + offset[1] - halfsize,
+            halfsize*2.f,
+            halfsize*2.f
+        );
+    }
```
DaWelter, line 676: @DaWelter This is a change I'm not so sure about ... I might decide to revert it or make it an option. Depends on what works best for future networks.

DaWelter: @sthalik , from my side you can merge it.

PR #1562, opened 2022-11-01
tracker/nn: Let cmake script support a greater variety of ONNX installs

DaWelter:
Is it convenient for you @sthalik ?

I tested it on Linux with the onnx `onnxruntime-linux-x64-*.tgz ` release archive and on Windows x86 with the nuget package.


DaWelter:
I removed the `CMAKE_INSTALL_PREFIX` path because I saw no reason so search in there. .. I think that's what it was for. To search there, right?

sthalik:
LGTM.

PR #1563, opened 2022-11-01
tracker/nn: Tweaks, refactoring

DaWelter:
Hey @sthalik

it's about time I push some other changes I've been working on. Sorry I know this PR is a bit big ...

The new features with the filter stuff won't be used with the current model. I have yet to come up with a better model. I have to admit I must have messed up when preparing the final training code for the model because even with that original code I only get models which are somewhat less accurate ...

But anyway, I guess the fixes and the refactoring alone is nice. I might remove everything else again.

* Add rudimentary test for two functions .. maybe more in future
* Fix the rotation correction from vertical translation
* Move preview class to new files
* Move neural network model adapters to new files
* Add utility functions for opencv
* Query the model inputs/outputs by name to see what is available
* Supports outputs for standard deviation of the data distribution - What you get if you let your model output the full parameters of a gaussian distribution (depending on the inputs) and fit it with negative log likelihood loss.
* Disabled support for sequence models
* Add support for detection of eye open/close classification. Scale uncertainty estimate up if eyes closed
* Add a deadzone filter which activates if the model supports uncertainty quantification. The deadzone becomes larger the more uncertain the model/data are. This is mostly supposed to be useful to suppress large errors when the user blinks with the eyes
* Fix distance being twice of what it should have been

### DIFF
--- a/tracker-neuralnet/deadzone_filter.cpp
+++ b/tracker-neuralnet/deadzone_filter.cpp
@@ -1,0 +1,162 @@
+#include "deadzone_filter.h"
+#include "model_adapters.h"
+#include "opencv_contrib.h"
+#include "unscented_trafo.h"
+
+#include <opencv2/core/base.hpp>
+#include <opencv2/core/matx.hpp>
+#include <opencv2/core/quaternion.hpp>
+
+namespace neuralnet_tracker_ns
+{
+
+using namespace cvcontrib;
+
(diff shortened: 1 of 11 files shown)


sthalik:
@DaWelter, would you like a new build of opentrack to be released?

DaWelter:
There is currently no real need from my side. There are the fixes of course. Their impact on the usability of the tracker is quite small though.

PR #1571, opened 2022-11-10
Add functionality to use user specified versions of wine.

arki-v1:
I've added the ability for a user to point to their own wine executable. This will allow users who wish or are required to use custom wine distributions on non-steam games the ability to do so. The UI isn't neat currently as I've no prior experience with QT.


sthalik (changes requested):

Loki419420 (approved):
nice

Priton-CE:
Seems like this PR has gone cold.
The proposed changes have already been implemented in other PRs. I think this one can be closed now.

PR #1590, opened 2022-12-20
Fix initialization order issues in migration code

DaWelter:
Hi,

I think there are some issues. There is this migration thingy ...
grafik

It is initialized in the dynamic initialization phase and it accesses this `migration_thunks` vector you see me change in the PR.

The vector must be initialized, too, but to my knowledge this was not guaranteed to have happended. The order between different translation units would be up to the tooling. See https://en.cppreference.com/w/cpp/language/initialization, Sec Dynamic Initialization (3).

In my case, asserts in the windows debug CRT memory management functions were triggered. 

This patch seems to fix it. I replaced the global static vars with local static vars which would be initialized on first use. c++11 makes this thread safe, too. See https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables.

Cheers

### DIFF
--- a/migration/migration.cpp
+++ b/migration/migration.cpp
@@ -26,14 +26,25 @@ using namespace options::globals;
 
 namespace migrations::detail {
 
-static std::vector<mptr> migration_list;
-static std::vector<mfun> migration_thunks;
+
+std::vector<mptr>& migrator::migration_list()
+{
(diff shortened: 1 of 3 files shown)


sthalik:
Looks good, thanks!

PR #1603, opened 2023-01-23
Create profiles dir using proven method

dlpierce:
Attempting to directly create the profiles directory is causing problems when packaging for Arch Linux. This fix will copy the directory from the srcdir using already established install functions.

See https://aur.archlinux.org/packages/opentrack-git#comment-897814


sthalik:
Fixed by a slightly different patch. Thanks.

PR #1619, opened 2023-02-26
proto-wine UI changes

Priton-CE:
see RFC https://github.com/opentrack/opentrack/issues/1618

PR #1620, opened 2023-02-26
RFC: proto-wine UI changes second attempt

Priton-CE:
new branch
old [PR](https://github.com/opentrack/opentrack/pull/1619)

This only contains the changes done to the UI of the proto-wine module.

### DIFF
--- a/proto-wine/ftnoir_protocol_wine.cpp
+++ b/proto-wine/ftnoir_protocol_wine.cpp
@@ -44,6 +44,7 @@ void wine::pose(const double *headpose, const double*)
 #ifndef OTR_WINE_NO_WRAPPER
         if (shm->gameid != gameid)
         {
+            qDebug() << "proto/wine: looking up gameData";
             QString gamename;
             QMutexLocker foo(&game_name_mutex);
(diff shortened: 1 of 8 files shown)


sthalik:
Merged, thanks. I have trivial changes but there's no point in holding up this PR.

sthalik:
@ljgdasfhk feel like giving this a spin?

ljgdasfhk:
> @ljgdasfhk feel like giving this a spin?

I do not mind.
But I would like to suggest the following:
1) make it possible to compile without ONNX (I don't have a video card)
2) sorry for the offtopic, this is what is really needed in battle: hotkeys for the selected axis values! (interface should be like curve setting)

sthalik:
> make it possible to compile without ONNX (I don't have a video card)

The only thing that requires onnxruntime is tracker/nn. It also doesn't use the CPU. It runs the network on one or more cores.

Are there any issues at this time?

ljgdasfhk:
> Are there any issues at this time?

i use point-tracker
i think everything works great

sthalik:
I fixed onnxruntime on 32-bit Linux and issues when starting tracking with only one camera present (more serious than it sounds).

PR #1627, opened 2023-03-20
Eyeware Beam using the C API

eyedav:


sthalik:
Oh. You were building as 64-bit. opentrack releases are 32-bit. This explains why non-mangled symbols don't resolve either way.

Issue #1559, opened 2022-10-25
Opentrack 2022.3.1 and .2 fails to build on Linux (Fedora 36)  with neuralnet tracker enabled.


XenoPL:
As per title attempt to build when path to onnxruntime is provided with error:

```
[ 84%] Building CXX object tracker-neuralnet/CMakeFiles/opentrack-tracker-neuralnet.dir/ftnoir_tracker_neuralnet.cpp.o
In file included from ~/tmp/opentrack/opentrack-opentrack-2022.3.2/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp:8:
~/tmp/opentrack/opentrack-opentrack-2022.3.2/tracker-neuralnet/ftnoir_tracker_neuralnet.h:30:10: fatal error: onnxruntime/core/session/onnxruntime_cxx_api.h: Nie ma takiego pliku ani katalogu
   30 | #include <onnxruntime/core/session/onnxruntime_cxx_api.h>
```

Opentrack 2022.3.0 compiles just fine. Diff on tracker-neuralnet/ftnoir_tracker_neuralnet.h file between 2022.3.0 and 2022.3.2:

```
diff ~/tmp/opentrack/opentrack-opentrack-2022.3.2/tracker-neuralnet/ftnoir_tracker_neuralnet.h ~/tmp/opentrack/opentrack-opentrack-2022.3.0/tracker-neuralnet/ftnoir_tracker_neuralnet.h 

30c30
< #include <onnxruntime/core/session/onnxruntime_cxx_api.h>
---
> #include <onnxruntime_cxx_api.h>

```
If I change the include back to the way it is in 2022.3.0 it goes a bit further but stops with another error:

```
~/tmp/opentrack/opentrack-opentrack-2022.3.2/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp: In constructor ‘neuralnet_tracker_ns::PoseEstimator::PoseEstimator(Ort::MemoryInfo&, Ort::Session&&)’:
~/tmp/opentrack/opentrack-opentrack-2022.3.2/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp:420:62: error: ‘struct Ort::Session’ has no member named ‘GetOutputName’
  420 |     if (session_.GetOutputCount() >= 3 && "box"s == session_.GetOutputName(2, allocator_))
      |                                                              ^~~~~~~~~~~~~
[...]

PR #1638, opened 2023-04-06
tracker/neuralnet: Add support for building on Linux.

zhaozhixu:
* Remove tracker-neuralnet cmake code that prevents building on Linux.
* Update onnxruntime calling code for its API changes. onnxruntime has changed GetInputName/GetOutputName to GetInputNameAllocated/GetOutputNameAllocated in API version 12. Add compiling macro tests for that.

Fix issue #1559.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ if(CMAKE_COMPILER_IS_GNUCXX AND NOT __otr_compile_flags_set)
     set(__otr_compile_flags_set TRUE CACHE INTERNAL "" FORCE)
     foreach(i C CXX)
         set(CMAKE_${i}_FLAGS_RELEASE "-O3 -march=native" CACHE STRING "" FORCE)
-        set(CMAKE_${i}_FLAGS "-ggdb -Wall -Wextra -Wpedantic -fopenmp" CACHE STRING "" FORCE)
+        set(CMAKE_${i}_FLAGS "-ggdb -Wall -Wextra -Wpedantic" CACHE STRING "" FORCE)
     endforeach()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ if(CMAKE_COMPILER_IS_GNUCXX AND NOT __otr_compile_flags_set)
     set(__otr_compile_flags_set TRUE CACHE INTERNAL "" FORCE)
     foreach(i C CXX)
         set(CMAKE_${i}_FLAGS_RELEASE "-O3 -march=native" CACHE STRING "" FORCE)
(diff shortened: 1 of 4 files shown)


tracker-neuralnet/CMakeLists.txt
```diff
@@ -1,8 +1,5 @@
 include(opentrack-opencv)
 set(host-spec "${CMAKE_SYSTEM_NAME} ${CMAKE_SYSTEM_PROCESSOR} ${CMAKE_SIZEOF_VOID_P}")
-if(host-spec MATCHES "^Linux i[3-6]86 4$")
```
sthalik, line 3: This shouldn't be necessary.

See https://github.com/opentrack/opentrack/blob/master/cmake/FindONNXRuntime.cmake#L36

CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ if(CMAKE_COMPILER_IS_GNUCXX AND NOT __otr_compile_flags_set)
     set(__otr_compile_flags_set TRUE CACHE INTERNAL "" FORCE)
     foreach(i C CXX)
         set(CMAKE_${i}_FLAGS_RELEASE "-O3 -march=native" CACHE STRING "" FORCE)
-        set(CMAKE_${i}_FLAGS "-ggdb -Wall -Wextra -Wpedantic" CACHE STRING "" FORCE)
+        set(CMAKE_${i}_FLAGS "-ggdb -Wall -Wextra -Wpedantic -fopenmp" CACHE STRING "" FORCE)
```
sthalik, line 57: Don't touch the toplevel makefile.

tracker-neuralnet/model_adapters.cpp
```diff
@@ -310,13 +328,18 @@ std::optional<PoseEstimator::Face> PoseEstimator::run(
 
     try
     {
+        std::vector<const char *> input_names, output_names;
+        std::transform(input_names_.begin(), input_names_.end(), std::back_inserter(input_names),
```
sthalik, line 332: Don't allocate memory in the main loop.

sthalik (changes requested):

tracker-neuralnet/CMakeLists.txt
zhaozhixu: Yes, removed in latest patch.

CMakeLists.txt
zhaozhixu: OK. Done in cmake of neuralnet.

tracker-neuralnet/model_adapters.cpp
zhaozhixu: Fixed by adding a vector member for C strings.

DaWelter:
Oh I guess I'm a bit late. Regardless. Nice work.
But are you sure you need to copy the names into `std::string`?

PR #1665, opened 2023-05-28
Improved chroma key for point tracker

tombrazier:
I have been trying to use chroma key to isolate the light from green LEDs on my headset clip. However the middle of the LEDs is so bright that it is overexposed on the camera and the color fades to white. i.e. I get white blobs with green halos. The chroma key logic sees this as dark blobs surrounded by light halos. This confuses the subsequent thresholding logic. I can turn down the camera exposure but this results in very small blobs and these decrease point tracker's accuracy.

This PR adds an option for chroma key to detect overexposed pixels as the requested color when using any of the chroma key colors. This works really well.

Using green as an example, the logic change is...

Existing logic: every pixel is transformed to a grayscale value with the formula `G - 0.5 * (R + B)`. (Where `R`, `G` and `B` and the red, green and blue channel intensities.) Subsequently a threshold is applied. The greener a pixel, the more likely the transformed value is to exceed the threshold.

New logic: The green channel intensity is used to calculate a weighted average between the green channel and the grayscale value calculated above. So the new formula is `G/255 * G + (1 - G/255) * (G - 0.5 * (R + B))`. This simplifies to `G - 0.5 * (R + B) + G/255 * 0.5 * (R + B)`, which is the existing chroma grayscale value plus `G/255 * 0.5 * (R + B)`.

With the new logic, the closer G is to 255, the less color saturation is expected from the pixel.

I have also added a checkbox to the options which is enabled when chroma key is selected. This allows either the new or the existing logic to be selected. It defaults to the existing logic so that a new version of opentrack would not change the behaviour without the user specifically requesting it.

### DIFF
--- a/tracker-pt/module/point_extractor.cpp
+++ b/tracker-pt/module/point_extractor.cpp
@@ -11,6 +11,7 @@
 #include "frame.hpp"
 #include "cv/numeric.hpp"
 #include "compat/math.hpp"
+#include "compat/math-imports.hpp"
 
 #include <opencv2/imgproc.hpp>
 
@@ -115,20 +116,26 @@ void PointExtractor::filter_single_channel(const cv::Mat& orig_frame, float r, f
 {
     ensure_channel_buffers(orig_frame);
 
-    cv::transform(orig_frame, dest, cv::Mat(cv::Matx13f(b, g, r)));
-
-    // decrease color saturation logic in proportion to key color intensity?
-    if (overexp)
+    // just filter for colour or also include overexposed regions?
+    if (!overexp)
+        cv::transform(orig_frame, dest, cv::Mat(cv::Matx13f(b, g, r)));
+    else
     {
-        // get the intensity of the key color (i.e. +ve coefficients)
-        cv::Mat1b temp1(orig_frame.rows, orig_frame.cols);
-        cv::transform(orig_frame, temp1, cv::Mat(cv::Matx13f(std::max(b, 0.0f), std::max(g, 0.0f), std::max(r, 0.0f))));
-        // get the intensity of the non-key color (i.e. -ve coefficients)
-        cv::Mat1b temp2(orig_frame.rows, orig_frame.cols);
-        cv::transform(orig_frame, temp2, cv::Mat(cv::Matx13f(std::max(-b, 0.0f), std::max(-g, 0.0f), std::max(-r, 0.0f))));
-        // add non-key color back into the result in proportion to key color intensity
-        cv::multiply(temp1, temp2, temp1, 1.0 / 255);
-        cv::add(dest, temp1, dest);
+        for (int i = 0; i < orig_frame.rows; i++)
+        {
+            cv::Vec3b const* const __restrict orig_ptr = orig_frame.ptr<cv::Vec3b>(i);
+            uint8_t* const __restrict dest_ptr = dest.ptr(i);
+            for (int j = 0; j < orig_frame.cols; j++)
+            {
+                // get the intensity of the key color (i.e. +ve coefficients)
+                uchar blue = orig_ptr[j][0], green = orig_ptr[j][1], red = orig_ptr[j][2];
+                float key = std::max(b, 0.0f) * blue + std::max(g, 0.0f) * green + std::max(r, 0.0f) * red;
+                // get the intensity of the non-key color (i.e. -ve coefficients)
+                float nonkey = std::max(-b, 0.0f) * blue + std::max(-g, 0.0f) * green + std::max(-r, 0.0f) * red;
+                // the result is key color minus non-key color inversely weighted by key colour intensity
+                dest_ptr[j] = std::max(0.0f, std::min(255.0f, key - (255.0f - key) / 255.0f * nonkey));
+            }
+        }
     }
 }
 
@@ -166,7 +173,6 @@ void PointExtractor::color_to_grayscale(const cv::Mat& frame, cv::Mat1b& output)
     case pt_color_green_chromakey:
     {
         filter_single_channel(frame, -0.5, 1, -0.5, s.chroma_key_overexposed, output);
-
         break;
     }
     case pt_color_blue_chromakey:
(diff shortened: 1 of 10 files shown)


sthalik:
The original chroma key support was written by https://github.com/a1k0n. He should weight in if he wants. If he doesn't, I'm going to assume that he agreed, and reviewing this pull request.

tombrazier:
I have implemented a performance improvement. Are you okay if I push further changes?

Also, it occurs to me that it could be helpful to introduce a new slider to specify how narrow the chroma key band is. (At least it would be useful to me!) This would work with or without the new overexposed logic, but would be especially beneficial in further tuning the case with overexposure. I presently have whitish backgrounds in my room which, depending on ambient light levels, still pass the threshold because they are bright enough. I have experimented and found that a tighter definition of "green" for lower light level areas fixes this. If others think such a new slider would be helpful I could bundle with with this PR or submit another.

Thoughts?

sthalik:
Feel free to push into the PR's branch.

It would help with testing and debugging if you made some sample videos demonstrating color key tracking, that can be replayed using virtual webcam software.

tombrazier:
Okay, change pushed.

The replay idea is brilliant. Will record some videos.

tombrazier:
Test video:

https://github.com/opentrack/opentrack/assets/68918209/61cb37d1-7e99-4d57-ba39-4e74dee3430a

The shiny highlights on the door in the background present a fun challenge.

GO63-samara:
I can't see the video.

tombrazier:
Doesn't play in my web browser either. Plays fine on VLC, though. Must be some codec thing. Hang on...

tombrazier:
https://github.com/opentrack/opentrack/assets/68918209/572bff50-2839-4a12-9f8d-4de277901f11

Right, different video format.

sthalik:
FWIW, this software allows video input without having to pirate it or run as a trial version. Be sure to install the 32-bit version.

- https://github.com/webcamoid/akvirtualcamera/releases
- https://github.com/webcamoid/webcamoid/releases/tag/daily-build

When adding the output method, don't specify a format but add it directly (or else output won't work). You might get a Vulkan-related error on startup which can be safely ignored. The joys of open-source.

tombrazier:
> FWIW, this software allows video input without having to pirate it or run as a trial version. Be sure to install the 32-bit version.

Thanks.

I have added the slider for "chroma key strength" now and I reckon I have done all I set out to do with this PR (potential bugs, etc. aside of course!).

tracker-pt/module/point_extractor.cpp
```diff
@@ -111,11 +112,31 @@ void PointExtractor::extract_single_channel(const cv::Mat& orig_frame, int idx,
...
+    {
+        for (int i = 0; i < orig_frame.rows; i++)
+        {
+            cv::Vec3b const* const __restrict orig_ptr = orig_frame.ptr<cv::Vec3b>(i);
+            uint8_t* const __restrict dest_ptr = dest.ptr(i);
+            for (int j = 0; j < orig_frame.cols; j++)
+            {
+                // get the intensity of the key color (i.e. +ve coefficients)
+                uchar blue = orig_ptr[j][0], green = orig_ptr[j][1], red = orig_ptr[j][2];
+                float key = std::max(b, 0.0f) * blue + std::max(g, 0.0f) * green + std::max(r, 0.0f) * red;
```
sthalik, line 132: You want `std::fmaxf` and `std::fminf`.

tracker-pt/module/point_extractor.cpp
```diff
@@ -146,32 +167,38 @@ void PointExtractor::color_to_grayscale(const cv::Mat& frame, cv::Mat1b& output)
     }
     case pt_color_red_chromakey:
     {
-        filter_single_channel(frame, 1, -0.5, -0.5, output);
+        float non_key_coeff = -0.5 * *s.chroma_key_strength;
```
sthalik, line 170: You can hoist `non_key_coeff` out of the switch.

tracker-pt/module/point_extractor.cpp
```diff
@@ -146,32 +167,38 @@ void PointExtractor::color_to_grayscale(const cv::Mat& frame, cv::Mat1b& output)
...
     {
-        filter_single_channel(frame, -0.5, -0.5, 1, output);
+        float non_key_coeff = -0.5 * *s.chroma_key_strength;
+        filter_single_channel(frame, non_key_coeff, non_key_coeff, 1, s.chroma_key_overexposed, output);
         break;
     }
     case pt_color_cyan_chromakey:
     {
-        filter_single_channel(frame, -1, 0.5, 0.5, output);
+        float non_key_coeff = -1.0 * *s.chroma_key_strength;
```
sthalik, line 188: and to save on typing, replace this with `2 * non_key_coeff`.

sthalik:
In the UI on the model tab you inadvertedly switched the tab to the custom model.

sthalik:
Edit the widget tab order and fix the new chroma key options appearing out of order.

tracker-pt/module/point_extractor.cpp
tombrazier: std::fmax was a whole load slower than std::max in a debug build. I'll have a look and see how it compares in release.

tracker-pt/module/point_extractor.cpp
tombrazier: Okay, will do.

tracker-pt/module/point_extractor.cpp
tombrazier: I think it's a bit less readable that way but it's your project. I'll do what you prefer.

tombrazier:
> In the UI on the model tab you inadvertedly switched the tab to the custom model.

Thanks.

> Edit the widget tab order and fix the new chroma key options appearing out of order.

Would you believe I already spent ages on trying to get tab order right? Anyway I am fairly confident about it now.

sthalik:
> On Win32, std::max() compiles to the single opcode MAXSS whereas std::fmax() calls a C library function which is a lot more expensive. Basically, std::max lets the compiler choose the best optimisation whereas std::fmax() doesn't. Is there a reason to prefer std::fmax()?

This shouldn't be happening.

There's nothing left for me to complain about. Let's merge this.

tombrazier:
> There's nothing left for me to complain about. Let's merge this.

Fab. Thanks.

PR #1666, opened 2023-05-31
"Natural Movement" Filter

tombrazier:
This is a new filter which is both very responsive when you move your head intentionally and also pretty much dead still when you don't. It gives me noticably better results than I am able to get from the other filters.

Having got the the point where I really like it, I realise I have converged on something very similar to EWMA, but with a few key differences. I am raising this PR but I suspect it would be better to extend EWMA instead. (And it would mean I don't have to decide whether I like "natural movement" as a name for the filter. Picking a name was the hardest part of the development!) It would be helpful @sthalik, et al to know what you'd prefer me to do. There is info below about how I think EWMA could be extended.

### Here's my blurb about this "natural movement" filter

The natural movement of our heads is different to what we think is happening and our brains are good at filtering out the difference. This filter attempts to do the same. It assumes the following:
1. When our head is moving at speed, it doesn't really need to track exactly. It just needs to track reasonably smoothly and end up at approximately the right place.
1. When we stop moving our head there is a slight natural "rebound" which we do not even notice, but the tracker does.
1. When our head is "still" it isn't really still: there is a combination of mid frequency vibrations called "[micro movements](https://matrix-wellbeing.iten-translations.com/the-vital-role-of-the-muscles-micro-movements/)" and slow position drift that we do not even realise is happening.
1. There is also higher frequency  sampling noise from the tracker.

To make movement feel natural, this filter applies some limited smoothing to the average speed of movement. When the speed is high (e.g. when glancing around) it tracks movement with the same level of smoothing as was applied to the speed. (This deals with 1. and 2.) When the speed is quite low (i.e. when we think we are actually sitting still), the filtering becomes so extreme that the filter output position drifts away from the input, resulting in an almost stationary output. (This deals with 3. and 4.) In between these extremes, the filter scales its behaviour smoothly.

The net effect is like having a really large deadzone and also a really high responsiveness and also general smooth movement.

### How this differs from EWMA

- Natural Movement (let's just call it NM) bases its behaviour on average speed. EWMA bases its behaviour on the standard deviation of input position vs output position. This means when your head feels still but is really drifting slightly EWMA will keep tracking the drift once it goes beyond a certain threshold. NM pretty much stops tracking the drift at all (but does catch up any time you make a faster head movement).
- EWMA automatically tunes its noise sensitivity whereas NM has it set statically via its dialog. I quite like EWMAs approach and would consider borrowing it for NM.
- EWMA has a setting for the smoothing level when your head is still. NM just causes the movement to cease when your head is still. Because of how the smoothing is specified for EWMA it isn't actually possible to specify this level of smoothing - the config value would have to be infinite.
- Related to the previous point: The way EWMA and NM scale their smoothing between high and low speed conditions is different. Broadly speaking, there is an inverse linear (i.e. hyperbolic) relationship between the two scales. I find that the NM approach works better.

### How EWMA could be extended

- I think two settings could be added to the EWMA config: 1. a drop-down for "Use position" and "Use speed" and 2. a check-box for "Alternative smoothing scale".
- Use position: is the existing EWMA behaviour for calculating std. deviation of distance between input and output positions. This would be the default.
- Use speed: is the NM logic which uses smoothed speed instead of std. deviation of position delta.
- Alternative smoothing scale: would apply the transition between slow and fast behaviour the NM way. By default the box would be unchecked. If this box is checked then the "Min" slider becomes disabled.
- Perhaps we need a note on the config screen saying that the new options may give better behaviour and are worth trying out.

Alternatively, there could be just one checkbox that allows all the new behaviour to be selected at once.


DaWelter:
Hi @tombrazier ,

thanks for your contribution. I tried it and found it really not bad. Even out of the box. Better than EWA.

Though NW got my view stuck a little too early which I suppose could have been improved by tweeking the settings.

EWA doeesn't set the bar though. Accela and Hamilton filter are the best. Their deadzone feature helps a lot.

filter-nm/ftnoir_filter_nm.cpp
```diff
@@ -0,0 +1,48 @@
...
+    else
+    {
+        const double dt = t.elapsed_seconds();
+        t.start();
+
+        for (int i = 0; i < 6; i++)
+        {
+            double speed = (input[i] - last_input[i]) / dt;
+            speeds[i] += dt * (double)s.responsiveness[i] * (speed - speeds[i]);
+            filtered_output[i] += dt * (double)s.responsiveness[i] * min(1.0, abs(speeds[i]) / (double)s.drift_speeds[i]) * (input[i] - filtered_output[i]);
```
DaWelter, line 40: Both, in line 39 and line 40, when `dt * responsiveness` becomes large, the output would overshoot. Possibly it could become unstable and explode. Not sure.
I recognize you have two EWA going on there. So consider limiting the smoothing factors `alpha`, e.g. shown for the speed here, by
```
timescale = 1./responsiveness;
alpha = dt / (dt + timescale);
speeds[i] += alpha * (speed - speeds[i]); // EWA
```
In line 40, I think it would be fine then to just replace `dt * (double)s.responsiveness[i]` by `alpha`.

tombrazier: Good point. In my head dt was always going to be very small but at a max value of 20 for `responsiveness` and below 20Hz camera frame rate that's going to be a problem. Your suggested EWA approach is spot on. I'll do that.

tombrazier:
> Though NW got my view stuck a little too early which I suppose could have been improved by tweeking the settings.

Yeah, tweak away.

> Accela and Hamilton filter are the best. Their deadzone feature helps a lot.

I used Accela for a long time but it still didn't give me what I wanted. The best I got was still a bit slow in responding to fast movements and the deadzone always either ended up too small to suit my head's drift or too large for small intentional head movements. A lot of this, of course, relates to personal taste.

I'd be interested in your thoughts on NM vs Accela or Hamilton.

GO63-samara:
I like your analysis of the Hamilton filter. I feel like I should answer.

> pos_max + pos_deadzone looks suspicious to me - it would make more sense it if were just pos_max.

If just `pos_max`, without `pos_deadzone`, then we get a stationary output until the input crosses the border of the dead zone, then the output jumps to the value of the dead zone.
The output must always not exceed the input by the amount of the dead zone. It's like if the input is inside a round barrel, and the output is the center of this barrel. When the input rests against the wall of the barrel, the center of the barrel moves. That is, the output is always less than the input.

> And I don't understand alpha = alpha * (dist - pos_deadzone) / (dist + EPSILON) at all.

By design, this line should scale the reduced output by the amount of the dead zone. And it looks like you found a bug (or a typo). Now I see that by design this line should be like this:
`alpha = alpha * (pos_max - pos_deadzone) / (pos_max + EPSILON);
`
Of course, this also needs to be checked.

> I think the code should be:
> 
>     double alpha = (dist - pos_deadzone) / (pos_max + pos_deadzone + EPSILON);
>     alpha = std::min(1.0, std::max(0.0, alpha));
>     if (alpha > 0.0) alpha = pow(alpha, pos_pow);

To avoid a jump in the output when crossing the border of the dead zone (noticeable with low anti-aliasing), a fourth line is needed:

```
    double alpha = (dist - pos_deadzone) / (pos_max + pos_deadzone + EPSILON);
    alpha = std::min(1.0, std::max(0.0, alpha));
    alpha = pow(alpha, pos_pow);
    alpha = alpha * (pos_max - pos_deadzone) / (pos_max + EPSILON);
```
But in general you are right. By changing the rule for calculating alpha, you can get a filter on a quaternion with various characteristics.

tombrazier:
Thanks @GO63-samara. I haven't quite understood what you're saying but it's late here. There's a good chance my brain will get it right away in the morning!

In the meantime, I have noticed an error in what I wrote and have added an edit noting the error. It's in the bit where I said "I think the code should be...".

tombrazier:
Just to check we are on the same page, my understanding is:
- From `dist == 0` to `dist == pos_deadzone`: there is no movement, i.e. alpha is 0.0.
- From `dist - pos_deadzone == 0` to `dist - pos_deadzone == pos_max`: the value of `alpha` scales from `0.0` to `1.0` on some power curve, i.e. there is some level of smoothing.
- And above `dist - pos_deadzone == pos_max`: `alpha` is `1.0`, i.e. there is no smoothing.

`pos_deadzone` is not included within `pos_max` and `dist` has to be `>= pos_max + pos_deadzone` for there to be no smoothing.

GO63-samara:
Not certainly in that way.

`alpha` must not reach `1.0`.

The fact is that the output tends not to the input, but to the border of the dead zone. Then there will be no jump when crossing the border of the dead zone.

That is, the output tends to ensure that the input is inside the barrel. But it does not strive for the entrance to be exactly in the center of the barrel.
And the center of the barrel is the output.

Then on your last chart, the upper limit will not be `1.0`, but `(10-1)/10 = 0.9`.
And it will be right.

tombrazier:
It makes sense to me this morning. I had not spotted the dead zone sized gap at the top of the graph. So the code should be:
```
    double alpha = (dist - pos_deadzone) / (pos_max + EPSILON);
    alpha = std::min(1.0, std::max(0.0, alpha));
    alpha = pow(alpha, pos_pow);
    alpha = alpha * (pos_max - pos_deadzone) / (pos_max + EPSILON);
```

This gives:
Screenshot_2023-06-03_10-31-44

Changes from the original are:
### Line 1
`double alpha = (dist - pos_deadzone) / (pos_max + pos_deadzone + EPSILON)` changed to `double alpha = (dist - pos_deadzone) / (pos_max + EPSILON)`.

This is so that `alpha = 0.0` when `dist == pos_deadzone` and `alpha = 1.0` when `dist == pos_max + pos_deadzone`.

### Lines 2 and 3 (which become line 2)
For both the VC++ and GCC compilers, using the STL's `std::min` and `std::max` functions allows the compiler to optimise the code better than using the standard library `fmax` and `fmin` functions. If it is known that C++17 or higher will always be used, this could be simplified to use `std::clamp`.

### Line 5 (which becomes line 4)
This is your change @GO63-samara, which I now understand.

tombrazier:
@GO63-samara I just realised I missed out a change to line4. I have now added it as an edit to my last post.

GO63-samara:
Then could you open another PR fixing the Hamilton filter?

tombrazier:
> Then could you open another PR fixing the Hamilton filter?

Will do.

sthalik:
@tombrazier can you kick the CI such as by sending an empty commit or rebasing?

tombrazier:
Will do. I still have a few things to do on this PR but I really like it. I have been sidetracked recently by other projects.

sthalik:
Alright, it passes CI.

Are you satisfied with the code's current state for it to be merged?

tombrazier:
I guess it can. I am still not certain about the name - it sounds pretentious. But if you're happy with it, that's okay with me.

I also want to convert the logic to use QQuaternion, but that can be a separate PR.

And I am planning to borrow from EWMA's automatic noise calibration algorithm. But that can also be a separate PR.

sthalik:
EWMA gathers the noise threshold from regular ingame movements. You might find that it does the opposite of what you want. Maybe if you add a tuning mode that doesn't run all the time it can be useful.

> I guess it can. I am still not certain about the name - it sounds pretentious. But if you're happy with it, that's okay with me.

If you want to pick another name there's still time before the release. Feel free to send a PR. When the module eventually gets renamed, a migration will be added to point existing users to the new name.

varbex:
@tombrazier Just wanted to leave a feedback, I´ve been using this with great results on a 60fps camera, the best thing about it compared to other filters is how easy it is to get to a great result just following your instructions in the filter tab. Thank you very much for the work put into this filter. Ps. Natural Movent is a great name.

tombrazier:
@varbex Thank you. I really like it myself as well. But it is great to hear I am not the only one who is finding it useful.

tombrazier:
PS my other contribution in the same release improves chroma-key point selection by including overexposed pixels. I used to find I had to keep tweaking the point detection settings as ambient light changed. I no longer have to do that. I use green LEDs so chroma key is great for me. I don't know how it would work with IR, whether that mainly looks red once the IR filter is removed from the camera or whether it affects all three colour sensors.

Issue #804, opened 2018-06-21
Prepar3d V4 issues with simconnect


MathijsG:
Hi Sthalik,

I'm trying to get Opentrack working with Prepar3D V4.

With the latest patches and with the latest V4 SDK version: 4.2.21.24048

However none of the entries of the FSX Simconnect output protocol work. I even tried installing all the different version at the redist folder of Prepar3D (not all at once ofcourse), but I still get this very familiar error:

![image](https://user-images.githubusercontent.com/529552/41737390-74f23e8e-758f-11e8-9ba2-f7e814fd4865.png)

FaceTrackNoIR does work, it has a Simconnect V2 protocol build in, since a few releases back, however I get low framerates with that somehow.

Do you have any direction for me to dive in or something which can help?

Where freetrack (meaning the protocol) always works without a hassle in every game, the simconnect protocol sounds like a little hell :-P.

sthalik:
@MathijsG,

There's a workaround to get simconnect working in the last stable version. Copy the simconnect `.dll` into one with the same name but no `lib` prefix.

But better use the unstable version. It's almost always less buggy than the stable version.

sthalik:
Been fixed in unstable. Closing.

MathijsG:
Do you have the compiled unstabled versions somewhere? Can't seem to find em on /releases/

sthalik:
http://opentrack.misaki.pl

MathijsG:
Thanks, that did the trick!

sthalik:
I'll be releasing a new version soon. Can you help with last-minute tests?

MathijsG:
Yeah, guide me master.

MathijsG:
@sthalik specific things where I can help?

sthalik:
@MathijsG I'm releasing a new build as soon as I commit individual changes. Can you post in #778?

The changes are largely internal (not user-visible) but there are some good nuggets here and there.

There's some new fancy stuff in the first post's changelog too.

sthalik:
@MathijsG see https://github.com/opentrack/opentrack/issues/778#issuecomment-400865602

sthalik:
I really should test this PT feature more. @MathijsG  Can you see if it helps reduce the amount of noise, or is it worse (more filtering needed)?

JDTiger:
Hey, i donwloaded the latest build from the your website, but I still get the simConnect error trying to use it with P3Dv4

MathijsG:
[...]

PR #1667, opened 2023-06-03
Hamilton documentation and minor fixes

tombrazier:
This PR began with some ideas about fixing the hamilton filter. After some discussion we concluded that there wasn't a great deal that needed fixing but some better documentation would be beneficial. This PR now addresses that along with a slight improvement in the `alpha` calculation logic. Also, whilst I was at it, I converted some tab indents to spaces to match the surrounding code.

# Start of the original PR...

See full discussion with @GO63-samara in #1666.

### Synopsis

As documented (in Russian but easily google translatable) [here](https://sites.google.com/site/diyheadtracking/home/opentrack/opentrack-hamilton-filter), the alpha value calculated for the hamilton filter should have the following curves:

image

It actually has this:

image

(where, in this case, dead zone is 1.0 and max distance is 10.0)

This PR the logic and the new curves are as intended:

image

### DIFF
--- a/filter-hamilton/ftnoir_filter_hamilton.cpp
+++ b/filter-hamilton/ftnoir_filter_hamilton.cpp
@@ -34,11 +34,15 @@ void hamilton::filter(const double *input, double *output)
     double dist	 = VectorDistance( &input[TX], pos_last);
 
     double alpha = (dist - pos_deadzone) / (pos_max + pos_deadzone + EPSILON);
-    alpha = fmin(alpha, 1.0);
-    alpha = fmax(alpha, 0.0);
-    alpha = pow (alpha, pos_pow);
-    alpha = alpha * (dist - pos_deadzone) / (dist + EPSILON);
-		
+    alpha = std::min(1.0, std::max(0.0, alpha));
+    if (alpha > 0.0)
+        alpha = pow(alpha, pos_pow);
+    // Scale alpha so that alpha * dist <= dist - pos_deadzone. This ensures that
+    // the center of the deadzone will never move closer to the input position than
+    // distance dist. And this ensures that the view never jumps ahead of head
+    // movements.
+    alpha *= (dist - pos_deadzone) / (dist + EPSILON);
+
     pos_last = Lerp(pos_last, input, alpha);
 
     output[TX] = pos_last.v[0]; 
@@ -62,10 +66,11 @@ void hamilton::filter(const double *input, double *output)
     double angle = AngleBetween(quat_input, quat_last);
 
     alpha = (angle - rot_deadzone) / (rot_max + rot_deadzone + EPSILON);
-    alpha = fmin(alpha, 1.0);
-    alpha = fmax(alpha, 0.0);
-    alpha = pow (alpha, rot_pow + rot_zoom);
-    alpha = alpha * (angle - rot_deadzone) / (angle + EPSILON);
+    alpha = std::min(1.0, std::max(0.0, alpha));
+    if (alpha > 0.0)
+        alpha = pow(alpha, rot_pow + rot_zoom);
+    // see comment in earlier alpha calculation above
+    alpha *= (angle - rot_deadzone) / (angle + EPSILON);
 
     quat_last = Slerp(quat_last, quat_input, alpha);
 
--- a/filter-hamilton/ftnoir_filter_hamilton.cpp
+++ b/filter-hamilton/ftnoir_filter_hamilton.cpp
@@ -52,11 +52,11 @@ void hamilton::filter(const double *input, double *output)
     // zoom smoothing:
     const double pow_zoom {s.kPowZoom};		
     const double max_z    {s.kMaxZ};
-		double rot_zoom = pow_zoom; 
+    double rot_zoom = pow_zoom;
 
     if (output[TZ] > 0) rot_zoom = 0;
-		else rot_zoom *= -output[TZ] / (max_z + EPSILON);
-		rot_zoom = fmin( rot_zoom, pow_zoom ); 
+    else rot_zoom *= -output[TZ] / (max_z + EPSILON);
+    rot_zoom = fmin( rot_zoom, pow_zoom );
 
     // rotations:
     const double rot_max     {s.kMaxRot};
--- a/filter-hamilton/ftnoir_filter_hamilton.h
+++ b/filter-hamilton/ftnoir_filter_hamilton.h
@@ -43,7 +43,7 @@ public:
     module_status initialize() override { return status_ok(); }
 private:
     tQuat   quat_last;
-	tVector pos_last;
+    tVector pos_last;
     settings s;
     bool first_run = true;
 };
--- a/filter-hamilton/ftnoir_filter_hamilton.h
+++ b/filter-hamilton/ftnoir_filter_hamilton.h
@@ -43,7 +43,7 @@ public:
     module_status initialize() override { return status_ok(); }
 private:
     tQuat   quat_last;
-		tVector pos_last;
+	tVector pos_last;
     settings s;
     bool first_run = true;
 };
--- a/filter-hamilton/ftnoir_hamilton_filtercontrols.ui
+++ b/filter-hamilton/ftnoir_hamilton_filtercontrols.ui
@@ -13,7 +13,7 @@
     <x>0</x>
     <y>0</y>
     <width>514</width>
-    <height>491</height>
+    <height>563</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -50,8 +50,8 @@
   <property name="styleSheet">
    <string notr="true"/>
   </property>
-  <layout class="QVBoxLayout" name="verticalLayout">
-   <item>
+  <layout class="QGridLayout" name="gridLayout_2">
+   <item row="2" column="0">
     <widget class="QGroupBox" name="groupBox">
      <property name="sizePolicy">
       <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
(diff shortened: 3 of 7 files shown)


GO63-samara:
Not sure about this fix:
`double alpha = (dist - pos_deadzone) / (pos_max + pos_deadzone + EPSILON);`
`double alpha = (dist - pos_deadzone) / (pos_max + EPSILON);`
Previously, with an increase in the dead zone, the smoothing zone also expanded. Now, when increasing the dead zone, the sum of the dead zones and smoothing will be constant == pos_max.
Try what is best. I don't know which is better.

`alpha *= (pos_max - pos_deadzone) / (pos_max + EPSILON);`
This is good. Maybe a little more readable:
`alpha *= 1.0 - pos_deadzone / (pos_max + EPSILON);`
One less variable, maybe a little faster?

GO63-samara:
And further.
I don't quite understand why you insist on including the line
`if (alpha > 0.0)`
Perhaps you want to exclude the `0^0` case? But `0^0=1`, so there's nothing wrong with that. Why extra `if`?

tombrazier:
@GO63-samara I'll look at your comments later. In the meantime, just a quick note to say I think your original last line might have been correct after all.

tombrazier:
> Not sure about this fix:
> `double alpha = (dist - pos_deadzone) / (pos_max + pos_deadzone + EPSILON);`
> `double alpha = (dist - pos_deadzone) / (pos_max + EPSILON);`
> Previously, with an increase in the dead zone, the smoothing zone also expanded. Now, when increasing the dead zone, the sum of the dead zones and smoothing will be constant == pos_max.

My assumption is that the sum of the dead zones and smoothing is where `alpha` becomes equal to 1.0. Or, putting it differently, `alpha == 1.0` when `dist = pos_max + pos_deadzone`. Is that correct? If so, try it out with the two formulea:

```
           1.0 = alpha = (dist - pos_deadzone) / (pos_max + pos_deadzone)
                   1.0 = (dist - pos_deadzone) / (pos_max + pos_deadzone)
pos_max + pos_deadzone = dist - pos_deadzone
                  dist = pos_max + 2 * pos_deadzone
```
vs
```
           1.0 = alpha = (dist - pos_deadzone) / pos_max
                   1.0 = (dist - pos_deadzone) / pos_max
               pos_max = dist - pos_deadzone
                  dist = pos_max + pos_deadzone
```
The second case is the one that matches my assumption.

> Try what is best. I don't know which is better.

I think either will work, but with different configuration. But only one of them matches your graph. And that depends on whether I have understood the graph correctly.

> alpha *= (pos_max - pos_deadzone) / (pos_max + EPSILON);
> This is good. Maybe a little more readable:
> alpha *= 1.0 - pos_deadzone / (pos_max + EPSILON);
> One less variable, maybe a little faster?

They are equivalent. As you say not referencing `pos_max` twice may be faster but I think there is overhead in loading 1.0 into a CPU register so they cost the same.

But I have been thinking: your original code would mean that when you do the linear interpolation there is never a case where the distance moved is `> dist - pos_deadzone` whereas with the new code if `dist > pos_deadzone + pos_max` then the resulting distance moved will be `> dist - pos_deadzone`. Am I right?

> I don't quite understand why you insist on including the line
> if (alpha > 0.0)
> Perhaps you want to exclude the 0^0 case? But 0^0=1, so there's nothing wrong with that. Why extra if?

0^0 == 1 is a problem for when `dist < pos_deadzone`. In that situation you want `alpha` to be 0.0. To be honest, though, 0^0 is a hypothetical situation because the smallest configurable value for `pos_pow` is 0.001.

I could actually just withdraw this PR. I don't think it really makes a huge difference to how the filter works. But it has been very helpful for me to understand the filter.

tombrazier:
Oh, I have worked out something that was confusing me: the New Input - Dead Zone line is not a horizontal line. The larger the value of New Input, the smaller Dead Zone should appear on the Y axis of this graph:

image

It really looks like this:
Screenshot_2023-06-04_15-13-38

GO63-samara:
`double alpha = (dist - pos_deadzone) / (pos_max + EPSILON);
`
It is not clear what will happen to alpha if `pos_max < pos_deadzone`.
I think there will be a problem.

But here there is no such problem, even if it does not fully correspond to the graph:
`double alpha = (dist - pos_deadzone) / (pos_max + pos_deadzone + EPSILON);
`

I agree with the rest.

DaWelter:
> `alpha *= (pos_max - pos_deadzone) / (pos_max + EPSILON);`
> This is good. Maybe a little more readable:

Are you sure. I can understand not wanting the previous version of this line. Now it's scaling the gain independent of the input based on how much `pos_max` is larger than `pos_deadzone`. Seems rather arbitrary. Maybe this line could be deleted entirely instead.

DaWelter:
I think of `pos_max` as the end of the transition zone where `alpha=1`. Therefore `pos_max` must be larger than `pos_deadzone`. Correct me if I'm wrong though.

GO63-samara:
`alpha` should not tend to `1.0`. This has already been discussed here: 
https://github.com/opentrack/opentrack/pull/1666#issuecomment-1574372634

tombrazier:
> It is not clear what will happen to alpha if pos_max < pos_deadzone.
> I think there will be a problem.

Here is a graph with `pos_max = 10.0` and `pos_deadzone = 20.0`:

Screenshot_2023-06-04_17-52-11

The existing behaviour is:

From `dist == 0.0` to `dist == pos_deadzone`: no movement.
From `dist >= pos_deadzone` to `dist == pos_max + 2 * pos_deadzone`: smoothing scales according to power curve.
From `dist >= pos_max + 2 * pos_deadzone`: minimum smoothing is applied.

My proposed behaviour is:

From `dist == 0.0` to `dist == pos_deadzone`: no movement.
From `dist >= pos_deadzone` to `dist == pos_max + pos_deadzone`: smoothing scales according to power curve.
From `dist >= pos_max + pos_deadzone`: minimum smoothing is applied.

However if I made this change it would mean existing users would have to reduce their `pos_max` values by `pos_deadzone`. I am happy just revert my change and stick with the existing behaviour.

I think I will add a comment to explain the last line.

Also, @GO63-samara what do you think about me putting some instructions into the configuration dialog? I suspect more people will use hamilton if they understand what the parameters do.

tombrazier:
> Seems rather arbitrary. Maybe this line could be deleted entirely instead.

@DaWelter The way @GO63-samara explained this line which made sense to me is imagine your head is inside a barrel. When you move your head to the edge of the barrel it begins to push the barrel to a new position. But in the new position your head is still touching the edge of the barrel - or maybe it is a little beyond the edge of the barrel and still pulling it into position. This imposes a limit of the value of `alpha`. It must not be that the new distance from the center of the barrel to your head is now less than the radius of the barrel. So when the `position += alpha * distance` linear interpolation happens, `alpha` must be small enough to ensure that the distance moved is not more than `dist - pos_deadzone`. The existing formula of `(dist - pos_deadzone) / dist` gives this guarantee.

> P.S. then there will also be no problems with alpha becoming negative.

I thought that too. Until I realised that the only way alpha could be multiplied by a negative number is if it is already 0.0.

> Finally, judging from double alpha = (dist - pos_deadzone) / (pos_max + EPSILON);, (which is fine imo), the pos_max behaves more like the size of the transition zone, starting after pos_deadzone, because alpha = 1 at dist == pos_deadzone + pos_max.

That is my understanding too.

GO63-samara:
> Also, @GO63-samara what do you think about me putting some instructions into the configuration dialog? I suspect more people will use hamilton if they understand what the parameters do.

Of course add some instructions.
If you are good with English, you can add the translation to https://github.com/opentrack/opentrack/wiki.
My English is not good enough.

tombrazier:
> If you are good with English, you can add the translation to https://github.com/opentrack/opentrack/wiki.

Do you know how to add a page to the wiki? It looks like you need write access. The google translation of your explanatory page is very good. We could just use that.

tombrazier:
@GO63-samara I have done as discussed and changed the title post for this PR.

The new config dialog looks like this:
hamilton

In addition to adding comments, I re-ordered the settings because I think it flows better to think in terms of dead zone &rarr; max distance &rarr; smoothing power. I'll change these back if you don't like it.

GO63-samara:
This is good

DaWelter:
> alpha should not tend to 1.0. This has already been discussed here
@GO63-samara It should be fine for alpha<1. Minus some wiggle room for rounding errors maybe. When alpha =1 the measured position is outputted. Correct me if I missed something.

> The existing formula of `(dist - pos_deadzone) / dist` gives this guarantee.
Hm ... yes ... I confused it earlier with something else. Now I see you reverted to this formula, and imo it's also good.

sthalik:
> @sthalik I don't have anything more I think needs doing on this and @GO63-samara is happy with it.

This time you guys have to review and merge it on your own. I can't concentrate on anything for the time being, sorry.

tombrazier:
> This time you guys have to review and merge it on your own. I can't concentrate on anything for the time being, sorry.

Sorry to hear that @sthalik. Does anyone else have write access to the repository?

DaWelter:
I can. As there are no more complaints ...

tombrazier:
Thank you. I have a few more contributions and then I'm done. @DaWelter are you happy if I continue making submissions? My remaining changes are:
- Much faster settling time for EWMA.
- CPU usage improvements for my chroma key changes.
- Efficient wait in the simconnect thread - although I have a question about that because sthalik removed the original 100ms wait in commit c26423c amongst a bunch of other changes. Those changes fixed P3D according to #804 but I don't have P3D to test on. But without the efficient wait that thread spins at 100% CPU.
- My NM filter with the speed based deadzone is maturing. It now uses @GO63-samara's quaternion and spherical deadzone logic. And I am planning to add a EWMA inspired automatic calibration option.

sthalik:
> But without the efficient wait that thread spins at 100% CPU.

The thread runs at 250 Hz. If a filter or proto can't run that fast, it's a bug.

Also look for stdout (need opening the app through windows console) [warning messages](https://github.com/opentrack/opentrack/blob/master/logic/pipeline.cpp#L596-L601) about the main loop being unable to keep up.

tombrazier:
> The thread runs at 250 Hz. If a filter or proto can't run that fast, it's a bug.

The `pipeline::run()` thread runs at 250Hz. I am referring to the `simconnect::run()` thread, which watches for `Frame` notifications from SimConnect. This thread sits in a tight loop calling `WaitForSingleObject()` and passing a timeout of 0ms. Change the timeout to 1ms and the CPU usage drops dramatically.

sthalik:
The code expects that `simconnect_calldispatch` doesn't return immediately, and succeeds each time while tracking.

Try looking at the timings, maybe there's a way to avoid context switches and jitter. But it definitely needs a fix one way or another.

tombrazier:
> Try looking at the timings, maybe there's a way to avoid context switches and jitter. But it definitely needs a fix one way or another.

So is the 0ms to avoid context switches? (i.e. by preventing the thread from going into an efficient wait state.)

`simconnect_calldispatch` does return immediately with FSX. I don't know about other simulators.

I have found (with FSX) that `simconnect_set6DOF` doesn't seem to mind which thread it is called from. So it can be called directly from `simconnect::pose`. With `simconnect::pose` being called 250 times a second, this should minimise jitter.

sthalik:
> So is the 0ms to avoid context switches? (i.e. by preventing the thread from going into an efficient wait state.)

If `calldispatch` didn't improperly return without blocking, that would be one context switch too many.

> I have found (with FSX) that simconnect_set6DOF doesn't seem to mind which thread it is called from.

Add a commit fixing it in preferred way for me to cherry-pick.

Possibly many users had the same problem and nobody said anything for years. Sad. It's good that you mentioned it, anyway.

tombrazier:
https://github.com/tombrazier/opentrack/tree/efficientwait is the simplest fix and it makes a big difference. But it would be good to have some users with P3D and MSFS test it.

GO63-samara:
@tombrazier 
One more note.
When I created the Hamilton filter, I did not yet know about the existence of the `QQuaternion` class in Qt.
If we rewrite the `hamilton-tools` files using the `QQuaternion` class, we get a shorthand for the Opentrack code.
You can also take into account that quaternions are used, for example, both in `neuralnet` and in `kinect-face` trackers.
This means that there are ways to optimize the code for other trackers too.

GO63-samara:
I also want to add that the `QQuaternion` class is already used in `pose-widget`.

tombrazier:
Thanks, I think this is an important observation. I will change all my code to use QQuaternion and, if you're happy with it, will also mod the Hamilton filter so there is consistency across all the modules.

PR #1671, opened 2023-06-13
Faster settle time for EWMA

tombrazier:
EWMA has a smoothing function with a time constant of 60s. This means that the filter has to run for a time on the order of 60s before it really starts to work.

This PR reduces the time constant when the filter starts so that it resolves a lot faster. Full accuracy will still take on the order of 60s but that is a marginal gain whereas an early estimate of the smoothed position value has a noticeable positive effect.

There are also a few minor code cleanups and efficiency improvements, including not applying the -180° to 180° logic for linear axes.


sthalik:
This makes sense, thanks.

sthalik:
If only because `full_turn` was applied to XYZ pose elements.

tombrazier:
> If only because `full_turn` was applied to XYZ pose elements.

Copy/paste error, presumably.

PR #1686, opened 2023-07-05
Migrate to procps-ng 4.0

berarma:
I've tried to make the changes needed for procps-ng 4.0. This is the first time I touch any procps related code. I'm not sure how good are the changes. Needs testing.

### DIFF
--- a/compat/process-list.hpp
+++ b/compat/process-list.hpp
@@ -139,24 +139,31 @@ QStringList get_all_executable_names()
 {
     QStringList ret;
     enum pids_item items[] = { PIDS_ID_PID, PIDS_CMD, PIDS_CMDLINE_V };
+
     enum rel_items { rel_pid, rel_cmd, rel_cmdline };
     struct pids_info *info = NULL;
     struct pids_stack *stack;
+    QString tmp; tmp.reserve(64);
 
     procps_pids_new(&info, items, 3);
 
-    while ((stack = procps_pids_get(info, PIDS_FETCH_TASKS_ONLY))) {
+    while ((stack = procps_pids_get(info, PIDS_FETCH_TASKS_ONLY)))
+    {
         char  **p_cmdline = PIDS_VAL(rel_cmdline, strv,  stack, info);
 
         // note, wine sets argv[0] so no parsing like in OSX case
-        if (p_cmdline && p_cmdline[0])
+        if (p_cmdline && p_cmdline[0] && p_cmdline[0][0] &&
+                !(p_cmdline[0][0] == '-' && !p_cmdline[0][1]))
         {
-            QString tmp(p_cmdline[0]);
+            tmp = QString{p_cmdline[0]};
             const int idx = std::max(tmp.lastIndexOf('\\'), tmp.lastIndexOf('/'));
-            tmp = tmp.mid(idx == -1 ? 0 : idx+1);
+            if (idx != -1)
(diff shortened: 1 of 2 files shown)


compat/process-list.hpp
```diff
@@ -131,33 +131,35 @@ static QStringList get_all_executable_names()
...
+            QString tmp(p_cmdline[0]);
             const int idx = std::max(tmp.lastIndexOf('\\'), tmp.lastIndexOf('/'));
             tmp = tmp.mid(idx == -1 ? 0 : idx+1);
             ret.append(tmp);
         }
-        freeproc(procs[i]);
     }
-    free(procs);
+
+    procps_pids_unref(&info);
```
sthalik, line 161: - Incorrect code style (curly brace must be on separate line, etc)
- Don't malloc twice in the loop
- On my Arch Linux it has a lot of fake elements with `"-"` as the contents.

Run the code below and see if it still works for you. Try uncommenting the `qDebug()` statements to be sure.

```c++
    enum pids_item items[] = { PIDS_ID_PID, PIDS_CMD, PIDS_CMDLINE_V };
    
    while ((stack = procps_pids_get(info, PIDS_FETCH_TASKS_ONLY)))
    {
        enum rel_items { rel_pid, rel_cmd, rel_cmdline };
        struct pids_info *info = NULL;
        struct pids_stack *stack;
        QString tmp; tmp.reserve(64);

        procps_pids_new(&info, items, 3);

        while ((stack = procps_pids_get(info, PIDS_FETCH_TASKS_ONLY)))
        {
            char  **p_cmdline = PIDS_VAL(rel_cmdline, strv,  stack, info);

            // note, wine sets argv[0] so no parsing like in OSX case
            if (p_cmdline && p_cmdline[0] && p_cmdline[0][0] &&
                !(p_cmdline[0][0] == '-' && !p_cmdline[0][1]))
            {
                tmp = QString{p_cmdline[0]};
                const int idx = std::max(tmp.lastIndexOf('\\'), tmp.lastIndexOf('/'));
                tmp = tmp.mid(idx == -1 ? 0 : idx+1);
                //qDebug() << "procps" << tmp;
                ret.append(tmp);
            }
    }
    //qDebug() << "-- procps end";

    procps_pids_unref(&info);
```

sthalik (changes requested):

compat/process-list.hpp
berarma: I've used your code removing a duplicated while statement at the start.

I'm testing on Debian 12 and there are no dashes, only valid commands. Could it be some security feature?

Thanks!

sthalik:
Merged, thanks.

sthalik:
In the future this code should use freelists with no mallocs at all but I don't have time right now. If it ain't broke don't fix it.

sthalik:
I can't find `libproc2` for Ubuntu `jammy` so for now I have to support both versions. Will fix it today.

berarma:
> I can't find `libproc2` for Ubuntu `jammy` so for now I have to support both versions. Will fix it today.

Yes, it seems Lunar is the first one that ships libproc2.

I would have liked to support both versions but I haven't enough experience with CMake/C++.

Thanks!

berarma:
> @berarma please test either `master` or [e050725](https://github.com/opentrack/opentrack/commit/e0507257a41c29fa2e9ac28f36470023a72c39c6). Thanks.

It's working with libproc2 on Debian 12.

PR #1705, opened 2023-08-25
fix build and key shortcuts on macOS + disable AppNap

matatata:
The qxt-mini library did not properly register keyboards presses and releases. I've fixed that. Without that the center binding would cause opentrack to always keep centering for instance. I also made small other tweaks to be able to build and run it.

### DIFF
--- a/macosx/make-app-bundle.sh
+++ b/macosx/make-app-bundle.sh
@@ -38,9 +38,9 @@ cp "$dir/PkgInfo" "$install/$APPNAME.app/Contents/"
 mkdir -p "$install/$APPNAME.app/Contents/MacOS/Plugins"
 cp -r "$install/Plugins" "$install/$APPNAME.app/Contents/MacOS/"
 
-# Add framework and other libraries and fixup other libraries
+# Use either of these, two of them at the same time will break things!
 macdeployqt "$install/$APPNAME.app" -libpath="$install/Library"
-sh "$dir/install-fail-tool" "$install/$APPNAME.app/Contents/Frameworks"
+#sh "$dir/install-fail-tool" "$install/$APPNAME.app/Contents/Frameworks"
 
 # Create an 512 resolution size for the icon (for retina displays mostly)
 #gm convert -size 512x512 "$dir/../gui/images/opentrack.png" "$tmp/opentrack.png"


sthalik:
I already cherry-picked the qxt-mini fix. It was a long time coming, thank you.

The macdeployqt change can't be merged, as Qt can be installed from anything from the official installer, macports, homebrew, or straight from the source code. So expecting the user to set up PATH is more of a general solution.

If the qApp assert fails, something's wrong. I guess it's before `main`? It should never be called like that. It becomes legal to call once QApplication [gets instantiated](https://github.com/opentrack/opentrack/blob/master/gui/init.cpp#L305). If it's called after it does and still crashes, then it can be something worse like multiple versions of Qt linked into the same process.

matatata:
Hi, I'm glad you could use some of my changes. In fact the keyboard change is the most important one. I'm not sure why the assert failed, but when I take it away the program works just fine.

There's one other important thing I noticed: macOS is throttling the process when in background rendering it somewhat useless. The feature is called **AppNap** and I've been looking into the docs and apparently you can write some code to let macOS know it should not throttle the process... such as https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/PrioritizeWorkAtTheAppLevel.html#//apple_ref/doc/uid/TP40013929-CH36-SW1

If I only knew how to add some objective-c code to the project I might be able to try things out. Do you have a clue where to put the code best? I'm struggling with all the make boilerplates and I just cannot figure out how to make the adaptations.

sthalik:
> If I only knew how to add some objective-c code to the project I might be able to try things out. Do you have a clue where to put the code best? 

Try [adding it](https://github.com/opentrack/opentrack/blob/master/cmake/opentrack-boilerplate.cmake#L47) after a space character.

```cmake
file(GLOB cxx CONFIGURE_DEPENDS "${dir}/*.cpp" "${dir}/*.m" "${dir}/*.mm")
```

The actual objc file extension notwithstanding.

opentrack/AppNapLib/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+    enable_language(OBJCXX)
+
+
+    add_library(AppNapLib SHARED
+        ${src_files}
+
+    )
+
+    target_link_libraries(AppNapLib
```
sthalik, line 18: With `STATIC`, try `target_link_libraries(AppNapLib INTERFACE "-framework Foundation")`.

opentrack/main-window.cpp
```diff
@@ -436,6 +441,11 @@ void main_window::start_tracker_()
     if (work)
         return;
 
+#ifdef __APPLE__
+    disable_appnap_start();
```
sthalik, line 445: LGTM 👍

opentrack/CMakeLists.txt
```diff
@@ -11,6 +11,11 @@ set_target_properties(opentrack-executable PROPERTIES
...
 
+
+if(APPLE)
+    add_subdirectory(AppNapLib)
+endif()
+
 set_source_files_properties(resources.rc OBJECT_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/opentrack.ico")
 
-target_link_libraries(${self} opentrack-user-interface opentrack-version)
+target_link_libraries(${self} opentrack-user-interface opentrack-version AppNapLib)
```
sthalik, line 21: Just add another `target_link_libraries()` line dependent on `APPLE`.

sthalik (changes requested):

sthalik:
@matata, See if this cleaned up diff works for you. If it does, I'm going to commit you with your name on it.

I don't have vmware unlocker installed right now so I can't run run the OSX build for the time being.

[patch.diff.txt](https://github.com/opentrack/opentrack/files/12441962/patch.diff.txt)

matatata:
> > If I only knew how to add some objective-c code to the project I might be able to try things out. Do you have a clue where to put the code best?
> 
> Try [adding it](https://github.com/opentrack/opentrack/blob/master/cmake/opentrack-boilerplate.cmake#L47) after a space character.
> 
> ```cmake
> file(GLOB cxx CONFIGURE_DEPENDS "${dir}/*.cpp" "${dir}/*.m" "${dir}/*.mm")
> ```
> 
> The actual objc file extension notwithstanding.

Thank you. That helped indeed. I've made the changes. It was quite simple in the end. It works fine so far.. no throttling any more. Please have a look.

matatata:
oh I just see that you had commented my previous attempt. As you can see I threw most of it away.

sthalik:
The cmake code still isn't ok. Just test this [patch](https://github.com/opentrack/opentrack/files/12441962/patch.diff.txt) via `git apply` on a clean repo and come back to me. This way we avoid a longer back-and-forth with me having to nitpick trivial lines in the commit.

matatata:
I'll work in your cleanups ..you sent me. sorry did not realise. was too focussed..

matatata:
I'm done so far. Unfortunately the *.mm change broke the linux build. Maybe you can help? I need to go now. Will check back later. CU

matatata:
Hi, 
the changes in your patches are good and do work except that I needed to add STATIC to the new target_sources directive in opentrack/CMakeLists.txt

I committed and pushed it to my fork:
https://github.com/opentrack/opentrack/commit/da767ee0039bd09914b57f3b99447749c9cf213f
I have also removed all  other commits that very indeed unnecessary, so you could cherry-pick that one app nap commit or use this patch I created.


diff --git a/opentrack/CMakeLists.txt b/opentrack/CMakeLists.txt
index 9bf5cb9b..0fd72475 100644
--- a/opentrack/CMakeLists.txt
+++ b/opentrack/CMakeLists.txt
@@ -13,4 +13,13 @@ set_target_properties(opentrack-executable PROPERTIES
 
 set_source_files_properties(resources.rc OBJECT_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/opentrack.ico")
 
+if(APPLE)
+    set_source_files_properties(appnap_mac.mm PROPERTIES COMPILE_FLAGS "-fno-objc-arc")
+    target_sources(${self} PRIVATE appnap_mac.mm)
+endif()
+
 target_link_libraries(${self} opentrack-user-interface opentrack-version)
+
+if(APPLE)
+    target_link_libraries(${self} "-framework Foundation")
+endif()
diff --git a/opentrack/appnap_mac.mm b/opentrack/appnap_mac.mm
new file mode 100644
index 00000000..3e0bea73
--- /dev/null
+++ b/opentrack/appnap_mac.mm
@@ -0,0 +1,47 @@
+#ifdef __APPLE__
+
+#import <Foundation/Foundation.h>
+
+/**
+ * Used to prevent macOS from throttling the opentrack process.
+ */
+
+id token = nil;
+
+void disable_appnap_start();
+void disable_appnap_stop();
+
+void disable_appnap_start() {
+
+    if(token){
+        NSLog(@"disable_appnap_start: already started");
+        return;
+    }
+
+
+    NSLog(@"disable_appnap_start");
+    token = [[NSProcessInfo processInfo]
+      beginActivityWithOptions: NSActivityUserInitiatedAllowingIdleSystemSleep
+      reason: @"Disable AppNap"];
+    [token retain];
+}
+
+void disable_appnap_stop() {
+    if(!token){
+        NSLog(@"disable_appnap_start: not started");
+        return;
+    }
+
+    NSLog(@"disable_appnap_stop");
+    [[NSProcessInfo processInfo] endActivity:token];
+    [token release];
+    token = nil;
+}
+
+
+
+#endif
+
+
+
+
diff --git a/opentrack/main-window.cpp b/opentrack/main-window.cpp
index f449845b..20506d81 100644
--- a/opentrack/main-window.cpp
+++ b/opentrack/main-window.cpp
@@ -32,6 +32,12 @@
 #include <QDir>
 #include <QDateTime>
 
+
+#ifdef __APPLE__
+void disable_appnap_start();
+void disable_appnap_stop();
+#endif
+
 extern "C" const char* const opentrack_version;
 
 using namespace options::globals;
@@ -436,6 +442,11 @@ void main_window::start_tracker_()
     if (work)
         return;
 
+#ifdef __APPLE__
+    disable_appnap_start();
+#endif
+
+
 #ifndef UI_NO_VIDEO_FEED
     auto* frame = ui.video_frame;
 #else
@@ -486,6 +497,10 @@ void main_window::stop_tracker_()
     if (!work)
         return;
 
+#ifdef __APPLE__
+    disable_appnap_stop();
+#endif
+
     force_is_visible(true);
     with_tracker_teardown sentinel;
 



Best regards

sthalik (approved):

sthalik:
Merged, thanks.

With regard to your `qApp` problem -- you may want to verify whether the assertion trips after QApplication has already been created. If it gets called before that, it's an opentrack bug. Otherwise:

```console
$ DYLD_PRINT_LIBRARIES=1 opentrack 2>&1 | grep Qt | sort
```

See if there are any duplicates, such as QtCore loaded from 2 different `.so`'s or something.

matatata:
Cool!

> With regard to your qApp problem -- you may want to verify whether the assertion trips after QApplication has already been created. If it gets called before that, it's an opentrack bug. Otherwise:
> $ DYLD_PRINT_LIBRARIES=1 opentrack 2>&1 | grep Qt | sort
> See if there are any duplicates, such as QtCore loaded from 2 different .so's or something

Yes it is very weird. I build like this:
git clone ...
cd opentrack
mkdir build
cd build
cmake .. -G "Unix Makefiles"
make install

# not3e that macqtdeploy does it’s thing as well...

Then I get a directory structure:

tree -L 1 install
install
├── Library
├── Plugins
├── doc
├── opentrack.app
└── thirdparty

I copy the opentrack.app to /Applications/

If I execute it and then there's the crash and your command lists multiple QT-Libs (I have qt installed via MacPorts.. perhaps that's a problem)

DYLD_PRINT_LIBRARIES=1 ./opentrack 2>&1 | grep Qt | sort
dyld[50124]: <231AA0A6-892A-3153-9FAD-772B0AE7794E> /Applications/opentrack.app/Contents/Frameworks/QtNetwork.framework/Versions/5/QtNetwork
dyld[50124]: <231AA0A6-892A-3153-9FAD-772B0AE7794E> /opt/local/libexec/qt5/lib/QtNetwork.framework/Versions/5/QtNetwork
dyld[50124]: <28FAEB0B-3490-390D-9AA5-84949F4C3F49> /Applications/opentrack.app/Contents/Frameworks/QtPrintSupport.framework/Versions/5/QtPrintSupport
dyld[50124]: <AD584864-61B5-3314-B217-BC71300B9430> /Applications/opentrack.app/Contents/Frameworks/QtDBus.framework/Versions/5/QtDBus
dyld[50124]: <AD584864-61B5-3314-B217-BC71300B9430> /opt/local/libexec/qt5/lib/QtDBus.framework/Versions/5/QtDBus
dyld[50124]: <CDF96ABB-CF78-394D-AECD-457451661250> /Applications/opentrack.app/Contents/Frameworks/QtCore.framework/Versions/5/QtCore
dyld[50124]: <CDF96ABB-CF78-394D-AECD-457451661250> /opt/local/libexec/qt5/lib/QtCore.framework/Versions/5/QtCore
dyld[50124]: <D9A77725-4BDE-34B6-BBDE-F8CD959AD9E6> /Applications/opentrack.app/Contents/Frameworks/QtGui.framework/Versions/5/QtGui
dyld[50124]: <D9A77725-4BDE-34B6-BBDE-F8CD959AD9E6> /opt/local/libexec/qt5/lib/QtGui.framework/Versions/5/QtGui
dyld[50124]: <F0EE5680-8A1E-3B78-ADA7-277C7EF6975D> /Applications/opentrack.app/Contents/Frameworks/QtMultimedia.framework/Versions/5/QtMultimedia
dyld[50124]: <F0EE5680-8A1E-3B78-ADA7-277C7EF6975D> /opt/local/libexec/qt5/lib/QtMultimedia.framework/Versions/5/QtMultimedia
dyld[50124]: <FEDF303E-E846-3328-A314-5A6FC53372BE> /Applications/opentrack.app/Contents/Frameworks/QtWidgets.framework/Versions/5/QtWidgets
dyld[50124]: <FEDF303E-E846-3328-A314-5A6FC53372BE> /opt/local/libexec/qt5/lib/QtWidgets.framework/Versions/5/QtWidgets
objc[50124]: Class KeyValueObserver is implemented in both /opt/local/libexec/qt5/lib/QtCore.framework/Versions/5/QtCore (0x10394a668) and /Applications/opentrack.app/Contents/Frameworks/QtCore.framework/Versions/5/QtCore (0x102cb0668). One of the two will be used. Which one is undefined.
objc[50124]: Class QMacAutoReleasePoolTracker is implemented in both /opt/local/libexec/qt5/lib/QtCore.framework/Versions/5/QtCore (0x10394a5c8) and /Applications/opentrack.app/Contents/Frameworks/QtCore.framework/Versions/5/QtCore (0x102cb05c8). One of the two will be used. Which one is undefined.
objc[50124]: Class QT_ROOT_LEVEL_POOL__THESE_OBJECTS_WILL_BE_RELEASED_WHEN_QAPP_GOES_OUT_OF_SCOPE is implemented in both /opt/local/libexec/qt5/lib/QtCore.framework/Versions/5/QtCore (0x10394a640) and /Applications/opentrack.app/Contents/Frameworks/QtCore.framework/Versions/5/QtCore (0x102cb0640). One of the two will be used. Which one is undefined.
objc[50124]: Class RunLoopModeTracker is implemented in both /opt/local/libexec/qt5/lib/QtCore.framework/Versions/5/QtCore (0x10394a6b8) and /Applications/opentrack.app/Contents/Frameworks/QtCore.framework/Versions/5/QtCore (0x102cb06b8). One of the two will be used. Which one is undefined.


When I then delete
├── Library
├── Plugin
from the install directory from above, then afterwards it starts fine and your command only shows:
DYLD_PRINT_LIBRARIES=1 ./opentrack 2>&1 | grep Qt | sort
dyld[50060]: <231AA0A6-892A-3153-9FAD-772B0AE7794E> /Applications/opentrack.app/Contents/Frameworks/QtNetwork.framework/Versions/5/QtNetwork
dyld[50060]: <28FAEB0B-3490-390D-9AA5-84949F4C3F49> /Applications/opentrack.app/Contents/Frameworks/QtPrintSupport.framework/Versions/5/QtPrintSupport
dyld[50060]: <4212A2A1-42D7-3F2B-AF56-CAB07BF467AF> /Applications/opentrack.app/Contents/Frameworks/QtSerialPort.framework/Versions/5/QtSerialPort
dyld[50060]: <AD584864-61B5-3314-B217-BC71300B9430> /Applications/opentrack.app/Contents/Frameworks/QtDBus.framework/Versions/5/QtDBus
dyld[50060]: <CDF96ABB-CF78-394D-AECD-457451661250> /Applications/opentrack.app/Contents/Frameworks/QtCore.framework/Versions/5/QtCore
dyld[50060]: <D9A77725-4BDE-34B6-BBDE-F8CD959AD9E6> /Applications/opentrack.app/Contents/Frameworks/QtGui.framework/Versions/5/QtGui
dyld[50060]: <F0EE5680-8A1E-3B78-ADA7-277C7EF6975D> /Applications/opentrack.app/Contents/Frameworks/QtMultimedia.framework/Versions/5/QtMultimedia
dyld[50060]: <FEDF303E-E846-3328-A314-5A6FC53372BE> /Applications/opentrack.app/Contents/Frameworks/QtWidgets.framework/Versions/5/QtWidgets

I don't get it. It’s magic.

Best regards

sthalik:
Then macdeployqt doesn't run `install_name_tool` properly. Consider modifying the build process to use the [old method](https://github.com/opentrack/opentrack/blob/master/macosx/install-fail-tool).

matatata:
> Then macdeployqt doesn't run `install_name_tool` properly. Consider modifying the build process to use the [old method](https://github.com/opentrack/opentrack/blob/master/macosx/install-fail-tool).

Actually I tried to remove the calls of that `install-fail-tool` script and at least for macdeployqt does very well. No issues. Perhaps it is not needed anymore and does more harm than good.

sthalik:
Just to make it clear, you no longer need to remove the `qApp` assert as long as `install-fail-tool` is commented out?

matatata:
Yes. Without the fail tool I can run the build without further action immediately after it is built from clean repo. the show dyn lib command shows no duplicatesAm 26.08.2023 um 17:53 schrieb Stanislaw Halik ***@***.***>:﻿
Just to make it clear, you no longer need to remove the qApp assert as long as install-fail-tool is commented out?

sthalik:
It's very important that this got fixed. Thanks for reporting that.

As for macdeployqt's path in the shell script -- generally for Unix systems users are expected to make a symlink to the binaries or simply add them to their PATH. Do you think this should be the case here or not?

sthalik:
There's also another osx issue where qxt-mini doesn't understand keystrokes such as 'home' or 'left arrow'. Since you knew how to fix the lack of the keyup event, you might want to look at that too.

matatata:
> It's very important that this got fixed. Thanks for reporting that.
> 
> As for macdeployqt's path in the shell script -- generally for Unix systems users are expected to make a symlink to the binaries or simply add them to their PATH. Do you think this should be the case here or not?

I did put it in the PATH eventually. You cannot know in advance where it will be on the user's system, so I think it's ok that the user compiling it needs to symlink it or put it in the PATH.

matatata:
> There's also another osx issue where qxt-mini doesn't understand keystrokes such as 'home' or 'left arrow'. Since you knew how to fix the lack of the keyup event, you might want to look at that too.

I just checked and seems all keys work except the keys on the numpad. Yes I plan to take a look at it in future. Out of curiosity: The Freetrack 2.0 Enhanced Output plugin is not available on Mac, because it's some third party thing, right?

sthalik:
>  The Freetrack 2.0 Enhanced Output plugin is not available on Mac, because it's some third party thing, right?

Use `proto/wine`. You must set a cache variable for it to become available.

PR #1730, opened 2023-09-30
tracker/nn: Fix compile error with old onnx version

DaWelter:
Hi @sthalik ,

just a small change. FYI, since I don't want to just push into upstream.

Cheers


sthalik:
Should there be an onnxruntime update before the next release?

DaWelter:
Update?

No. Not needed at all from my side. 

Tbh, this was a fix for my horribly outdated ONNX v1.6.

Issue #1728, opened 2023-09-30
[BUG] neuralNet roll is upside down


ddzzkunye:
neuralNet roll is upside down  
![example](https://github.com/opentrack/opentrack/assets/92249100/0f3309dd-61de-4266-b3ff-7c810c1a6bfa)

DaWelter:
I noticed and created a PR with a fix now. 

The visualization is actually working as intended, but the roll value was indeed by 180 deg wrong.

ddzzkunye:
Branches have been merged

PR #1731, opened 2023-09-30
tracker/nn: Fix 180 deg roll

DaWelter:
Hi @sthalik ,

this should fix the 180 degree roll offset of the NN tracker. (https://github.com/opentrack/opentrack/issues/1728)

Actually I've been playing with it based on an older master branch. So, unless you changed something related in the meantime, I'm pretty confident it'll do the trick.


sthalik:
Is this a regression or has it always been like this?

DaWelter:
I think it was always wrong :-/
And yes, I guess users will have to adjust their settings. :-/

ddzzkunye:
Just a negative symbol?

sthalik:
> I think it was always wrong

I was worried that it's a regression affecting multiple trackers. Good that it's not the case.

ddzzkunye:
aruco has a similar problem

PR #1735, opened 2023-10-01
tracker/nn: Add button to choose pose net file

DaWelter:
Hi,

this PR adds a button to the neuralnet tracker to chose the onnx file for the pose network.
It's meant for developers/me for more convenient experimentation because renaming and replacing the file is tiresome and error prone.

Screenshot at 2023-10-01 14-36-16

### DIFF
--- a/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
+++ b/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
@@ -27,6 +27,8 @@
 #include <QMutexLocker>
 #include <QDebug>
 #include <QFile>
+#include <QFileDialog>
+#include <QFileInfo>
 
 #include <cstdio>
 #include <cmath>
@@ -54,6 +56,12 @@ std::wstring convert(const QString &s) { return s.toStdWString(); }
 #else
 std::string convert(const QString &s) { return s.toStdString(); }
 #endif
+
+
+QDir get_default_model_directory()
+{
+    return QDir(OPENTRACK_BASE_PATH+ "/" OPENTRACK_LIBRARY_PATH "models");
+}
 
 
 int enum_to_fps(int value)
@@ -463,8 +471,7 @@ bool NeuralNetTracker::load_and_initialize_model()
 {
     const QString localizer_model_path_enc =
         OPENTRACK_BASE_PATH+"/" OPENTRACK_LIBRARY_PATH "/models/head-localizer.onnx";
-    const QString poseestimator_model_path_enc =
-        OPENTRACK_BASE_PATH+"/" OPENTRACK_LIBRARY_PATH "/models/head-pose.onnx";
+    const QString poseestimator_model_path_enc = get_posenet_filename();
 
     try
     {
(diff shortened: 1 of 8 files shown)


tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -900,6 +928,31 @@ void NeuralNetDialog::startstop_trans_calib(bool start)
...
+{
+    const auto root = get_default_model_directory();
+    // Start with the current setting
+    QString filename = settings_.posenet_file;
+    // If the filename is relative then assume that the file is located under the
+    // model directory. Under regular use this should always be the case.
+    if (QFileInfo(filename).isRelative())
+        filename = root.absoluteFilePath(filename);
+    filename = QFileDialog::getOpenFileName(this,
+        tr("Select Pose Net ONNX"), filename, tr("ONNX Files (*.onnx)"));
```
sthalik, line 941:
```suggestion
        tr("Select Pose Net ONNX"), filename, tr("ONNX files (*.onnx);;All files (*)"));
```

tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -900,6 +928,31 @@ void NeuralNetDialog::startstop_trans_calib(bool start)
...
+    if (filename.isEmpty())
+        return;
+    // When a file under the model directory was selected we can get rid of the
+    // directory prefix. This is more robust than storing absolute paths, e.g.
+    // in case the user moves the opentrack install folder / reuses old settings.
+    // When the file is not in the model directory, we have to use the absolute path,
+    // which is also fine as developer feature.
+    if (filename.startsWith(root.absolutePath()))
+        filename = root.relativeFilePath(filename);
+    settings_.posenet_file = filename;
```
sthalik, line 952: It's going to crash and burn with no user-facing indication when the file can't be opened for reading, right?

DaWelter:
Argh .. @sthalik , I pushed accidentally into upstream. Sorry, idk why my main branch was configured to track upstream ...

Anyhow. We can just continue here.

sthalik:
If you want to print an exception to the user, you can use `run_in_thread_async`. Just be sure to use a `[=]` capture for the QString.

tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
DaWelter: :sweat_smile: It was intended not to. But an exception got through.

Now tested with invalid file (the localizer onnx) and non-existing file. There will be no explicit error indication. The tracker just won't start. Ok?

sthalik: It's fine as-is. FWIW, you can out `run_in_thread_async` or `run_in_thread_sync`. They're a bit evil but in a good way.

sthalik: ...Or return `module_status` instead of `bool`.

DaWelter:
I would like to give users a new model or two to try.
Would be a good opportunity. But I'm not sure how to best approach it.

sthalik:
You mean a pre-release model? I can release a prerelease build with info on what's in it.

PR #1742, opened 2023-10-22
NeuralNet Tracker: New Models

DaWelter:
This PR adds two models to the tracker and sets git attributes to store them in git LFS.

**Changenotes**
* Trained on a larger, more diverse dataset which might transfer to better tracking quality across a wide user base IRL.
* Less twitching when user blinks, due to being trained on semi-synthetic examples with closed eyes.
* Supports the adaptive smoothing filter (built into the neuralnet-tracker) with the help of uncertainty estimation.
* Still unresolved is the pitch-yaw bias / systematic errors.
* Accuracy in terms of angle errors is on par with SOTA literature results from last year.
* The small model should run as fast as the original due to being of the same general architecture.
* The big model is ca 6x slower (on my machine) but less noisy.

The "0.2" in the filename refers to the release in the other repo https://github.com/opentrack/neuralnet-tracker-traincode/releases/tag/v0.2.

I wrote up my analysis results more thoroughly there:
https://github.com/opentrack/neuralnet-tracker-traincode/blob/develop/doc/v0.2-analysis/readme.md

**Usage**

Via the new file selector pick one of the `head-pose-*.onnx` files.
Screenshot at 2023-10-01 14-36-16


sthalik:
What does storing the binaries in GitHub LFS mean for users who may not necessarily have `git-lfs(1)` set up?

DaWelter:
> What does storing the binaries in GitHub LFS mean for users who may not necessarily have `git-lfs(1)` set up?

If not set up? Then users get "pointer files".

In my experience this can be a little bit annoying tbh. First, because the pointer files have the original filename. When trying to load them in the tracker they yield only a general error (in the console, currently).
Secondly, even with git lfs installed, sometimes the pointed to files don't get downloaded, requiring an extra `git lfs pull`. Or maybe that had to do with a particular "special" dev env ... idk. Still, I guess developers can deal with it.

Or do you suggest an alternate storage method?

Maybe we can see if it turns out to be a hassle?

sthalik:
> Or do you suggest an alternate storage method?

As long as they're below 100 MB they're technically capable of going into the repo.

As far as I'm concerned, you can just put them into the master branch. It's not like large files are updated more often than once every several months, and having them in branch history illustrates project's development over time. They also won't get accidentally lost since the master branch doesn't allow rewriting history (except in some emergency situations).

Alternatively, you can use a separate repo (such as `opentrack/opentrack-depends`) with an orphan branch named like `neuralnet-model/0.2`, then add it as a submodule. The reasoning is such that when you add another version (say, 0.3), you can delete the old branch and it doesn't stick around in the repo forever. You'd need a few lines of CMake `file(GLOB_RECURSE)` and `install(FILES)` handling to install all submodule `.onnx` files into the same install directory.

sthalik (approved):

DaWelter:
Thanks. As said in private, without git lfs the repository will contain all the files ever added in all version which will eventually add to the size a lot.

I tested it more now. Git clone from my github branch. Changing branches to this one and another one where the files don't exist. I was sceptical if the files would be restored correctly. But it worked. So far, no issues at all.

I guess it would be overkill to ever have more than a hand full of models in the repo. I expect to swap them out from time to time though. Hence, I'd say lfs makes sense.

DaWelter:
P.s. When the face crop is suitably chosen, the network can apparently surpass current SOTA angle prediction accuracy from research. And I'm *not* cheating, like training on the test data. :-)

| Model                    | Data                |   Yaw° |   Pitch° |   Roll° |   Mean° |   Geodesic° |   XY% |   S% |
|--------------------------|---------------------|--------|----------|---------|---------|-------------|-------|------|
| head-pose-0.2-small.onnx | AFLW 2k 3d / ROI1.2 |   4.51 |     2.81 |    2.75 |    3.36 |        5.47 |  2.78 | 3.37 |
| head-pose-0.2-small.onnx | AFLW 2k 3d / ROI1.0 |   4.36 |     2.70 |    2.43 |    **3.16** |        5.31 |  2.48 | 3.36 |

In spite of this, I find myself playing with the bigger network since it propagates less of input noise to the output.

rpgrunwald:
Could we please get an model specifically for eyeglass wearers?

DaWelter:
Request noted, but the short answer is probably not because I already picked the low hanging fruits and the other ones are hanging much higher.

How do the new models compare to the old one?

rpgrunwald:
Thanks for the response.   I can't speak to old vs. new as I just started using NeuralNet instead of AiTrack.  Subjectively, NeuralNet seems more responsive in Yaw but slightly less responsive in pitch.  CPU usage is noticeably lighter than AITrack (i7-8700K and I gave i2 threads).   I did not notice a great deal of performance difference between "head-pose-0.2-big.onnx" and "head-pose-0.2-small.onnx".  "head-pose-0.2-small.onnx" has much lower CPU utilization.  I'm using Accela filter.

What other information can I provide and how can I help?

PR #1813, opened 2024-03-09
Add DBus interfaces for controlling tracker

ananace:
I was finding some issues in properly binding the handling of the tracker on my local system, so I decided to hack together a quick set of DBus interfaces to make that particular thing easier to accomplish.

This exposes two interfaces on the session bus, `com.github.opentrack` which deals with starting/stopping the tracker, and - when the tracker is running - `com.github.opentrack.Tracker` which handles centering/enabling/zeroing the running tracker.

I had to expose the `set_center` function on the tracker pipeline, since DBus does best when working with momentary inputs. I also added some read-only properties to allow checking the states of toggles through the DBus interface.

### DIFF
--- a/cmake/opentrack-qt.cmake
+++ b/cmake/opentrack-qt.cmake
@@ -17,6 +17,14 @@ if(TRUE OR NOT APPLE AND NOT WINDOWS)
     endif()
 endif()
 
+set(OPENTRACK_DBUS_CONTROL CACHE BOOL "Expose DBus services for control purposes")
+if(OPENTRACK_DBUS_CONTROL)
+    add_definitions(-DOTR_DBUS_CONTROL)
+
+    list(APPEND qt-required-components "DBus")
+    list(APPEND qt-imported-targets Qt6::DBus)
+endif()
+
 find_package(Qt6 REQUIRED COMPONENTS ${qt-required-components} QUIET)
 find_package(Qt6 COMPONENTS ${qt-optional-components} QUIET)
 
--- a/logic/dbus.cpp
+++ b/logic/dbus.cpp
@@ -1,0 +1,28 @@
+#if defined OTR_DBUS_CONTROL
+#include "dbus.hpp"
+
+bool DBus::is_enabled() const
+{
+    return pipeline_->is_enabled();
+}
+
+bool DBus::is_zero() const
+{
+    return pipeline_->is_zero();
+}
+
+void DBus::Center()
+{
+    pipeline_->set_center(true);
+}
+
+void DBus::ToggleEnabled()
+{
+    pipeline_->toggle_enabled();
+}
+
+void DBus::ToggleZero()
+{
+    pipeline_->toggle_zero();
+}
+#endif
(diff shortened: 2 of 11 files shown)


sthalik:
We definitely need a way to remotely operate opentrack programmatically or from the command line, but dbus isn't the right solution for this. It's too platform-specific, has a lot of overhead, and integrating it is too invasive.

I'll implement it in a portable way using shared memory mappings (that can be used from opentrack's command-line or remotely with a stable ABI) once I finally take care of the issue affecting Windows releases.

mircobabini:
We are looking for the same thing, enable/disable open track from a nodejs application.
Is it something we can already achieve somehow? It's great to know @sthalik agreed on this topic.

mircobabini:
> I'll implement it in a portable way using shared memory mappings (that can be used from opentrack's command-line or remotely with a stable ABI) once I finally take care of the issue affecting Windows releases.

@sthalik do you have an ETA or maybe an approach so we can support you?

StelardActek:
@ananace Just wanted to let you know I've found your PR very useful. I patched it into my build of opentrack and added a button to my stream deck to recenter the tracking. I would love if something like this got official support.

asbytes:
@ananace Thanks for your PR. I assigned global hotkeys in Wayland DE (Plasma) without changing security settings for legacy X11 applications.

sthalik:
Conditionalize enabling D-Bus on `UNIX AND NOT APPLE`. Something like:

```cmake
set(opentrack-use-dbus 0)
if(UNIX AND NOT APPLE)
    set(OPENTRACK_DBUS CACHE ...)
    if(OPENTRACK_DBUS)
        set(opentrack-use-dbus 1)
        # enable all the Qt machinery
    endif()
endif()
```

Then in the module at the start:

```cmake
if(opentrack-use-dbus)
    otr_module(...)
endif()
```

Other than that it's time to merge it.

ananace:
Qt's D-Bus support is fully functional on both Windows and macOS - it's just an implementation of a protocol after all, so I really don't see a reason for forbidding it from being used there.

But default values makes a lot of sense OS-wise

sthalik:
That CMake code doesn't do what it should. Just drop the cmake code changes, I'll write them.

sthalik:
I've made some minor changes and amended the commit under your name. https://github.com/opentrack/opentrack/tree/feature/dbus

See if it's fine to commit it in the present shape.

ananace:
I'm still going to disagree with arbitrarily forbidding the D-Bus functionality from being used by Windows/macOS.  
Qt has full D-Bus support on those platforms too, including whole suites of applications that make use of it already, so if the user wants to build opentrack with D-Bus functionality on Windows/macOS then there's really no good reason to forbid it.

I'm not going to consider this limitation a hill to die on though, I personally only use Linux.

The changes seem fine.

mircobabini:
My colleagues and I are looking into this Dbus feature for Windows like 🥹

sthalik:
@ananace, you were right all along. 🤣

PR #1843, opened 2024-05-08
Combine Accela Filter with Hamilton Filter

DaWelter:
As promised a while ago, here is a filter which combines the Accela gain curves with the rotation-space based filtering from the Hamilton filter. I hope it could maybe replace both the Accela and Hamilton filters.

Basically it applies the accela gain to the rotation vector magnitude of the shortest rotation between the current orientation and the measured orientation.

The position calculation is like in the Accela Filter.

I also included a smoothing amplification depending on the z-position. In my version I add to the smoothing amount. In the Hamilton Filter, I think the gain curve was modified which I wouldn't know how to do in the Accela way.

Maybe I could delete some of the disabled code for a spline GUI. Then again, it might be useful for tweaking those gain curves to those who know what they are doing :-D

### DIFF
--- a/filter-accela-hamilton/accela_hamilton.cpp
+++ b/filter-accela-hamilton/accela_hamilton.cpp
@@ -26,7 +26,7 @@ accela_hamilton::accela_hamilton()
 
 void accela_hamilton::filter(const double* input, double *output)
 {
-    constexpr double EPSILON = 1e-30;
+    constexpr float EPSILON = 1e-15F;
 
     const QQuaternion current_rot = QQuaternion::fromEulerAngles(input[Pitch], input[Yaw], input[Roll]);
     const QVector3D current_pos(input[TX], input[TY], input[TZ]);
@@ -44,10 +44,10 @@ void accela_hamilton::filter(const double* input, double *output)
         return;
     }
 
-    const double pos_thres{s.pos_smoothing};
-    const double pos_dz{ s.pos_deadzone};
+    const float pos_thres{s.pos_smoothing};
+    const float pos_dz{ s.pos_deadzone};
 
-    const double dt = t.elapsed_seconds();
+    const float dt = t.elapsed_seconds();
     t.start();
 
     // Position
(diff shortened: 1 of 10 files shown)


GO63-samara:
Great job!
It would be useful to have the `s.max_z` point found automatically from the `Mapping Z` spline.
But I have no idea how to do this, maybe @sthalik can tell about this?

filter-accela-hamilton/accela_hamilton.cpp
```diff
@@ -0,0 +1,115 @@
...
+        output[TY] = output_pos.y();
+        output[TZ] = output_pos.z();
+        last_position = output_pos;
+    }
+
+    // Zoom smoothing:
+    const double max_zoomed_smoothing {s.max_zoomed_smoothing};		
+    const double max_z {s.max_z};
+
+    const double zoomed_smoothing = output[TZ]>0. ? 0. : std::min(1., -output[TZ] / (max_z + EPSILON))*max_zoomed_smoothing;
```
sthalik, line 70: Can you make it look less scary?

sthalik (approved):

sthalik:
Try explicitly clamping the position value to be between `1e-3` and `1e3`.

Should I come up with a `QQuaternion` replacement that uses double precision?

filter-accela-hamilton/accela_hamilton.cpp
DaWelter: Like it is now?

DaWelter:
> Try explicitly clamping the position value to be between `1e-3` and `1e3`.

What? Where?

> Should I come up with a `QQuaternion` replacement that uses double precision?

Nah. we can get way with float, I think. The mixing of float and double was awkward though. So I changed everything to float.

sthalik:
> What? Where?

Wasn't `z_max` an attempt at getting an upper bound on input Z values? You can just clamp them to a conservative range and remove the user-visible option.

filter-accela-hamilton/accela_hamilton.cpp
sthalik: Still don't like it, but it wouldn't be productive for me to keep nitpicking it. You know what you're doing.

DaWelter:
> > What? Where?
> 
> Wasn't `z_max` an attempt at getting an upper bound on input Z values? You can just clamp them to a conservative range and remove the user-visible option.

IC, no its the distance where the smoothing assumes the maximum value.
Poorly named I guess ...

filter-accela-hamilton/accela_hamilton.cpp
DaWelter: :-O

dbaarda:
I see I'm added as a reviewer on this. I'm super rusty on this code and probably not the right person to review this.

I had a glance... It looks OK to me, but I didn't look very closely.

DaWelter:
@dbaarda Thanks for taking a look even if superficial. sthalik and GO63-samara are available for reviews after all.

So let's just merge it.

PR #1845, opened 2024-05-09
tracker/nn: Add support for 3x3 pos/size cov matrix (and some cleanup)

DaWelter:
This PRs main purpose is to add support for 3x3 covariance matrices for position+size estimation.
My recent (unreleased) models output those instead of only component wise standard deviations.

Overall
- Accept 3x3 covariance matrix output for position+size
- Old models still worked fine in my testing
- Delete commented code which was there previously to support recurrent models. (Won't be used in foreseeable future)
- Delete leftover code for experiments with closed eyes detection (Likewise, I have given up on that / Is not really needed any more)
- Improve comments and code quality in the internal filter code

### DIFF
--- a/tracker-neuralnet/deadzone_filter.cpp
+++ b/tracker-neuralnet/deadzone_filter.cpp
@@ -3,6 +3,7 @@
 #include "opencv_contrib.h"
 #include "unscented_trafo.h"
 
+#include <tuple>
 #include <opencv2/core/base.hpp>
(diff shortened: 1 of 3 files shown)


sthalik (approved):
👍

sthalik:
As far as the unitialized metadata version field is concerned, could you add a custom metadata field to new models? Then the value would signify that the metadata isn't left uninitialized.

PR #1862, opened 2024-06-04
Prevent compile error when defining QT_NO_DEBUG_OUTPUT

Ayexor:
When definit ```QT_NO_DEBUG_OUTPUT```, the ```qDebug()``` macro expands to ```while (false) QMessageLogger().noDebug()```, which creates a syntax error when compined with the ```eval_once``` macro.

Editing the ```eval_once``` macro would be easier but I did not see a nice solution that works in all cases.

### DIFF
--- a/compat/check-visible.cpp
+++ b/compat/check-visible.cpp
@@ -70,7 +70,9 @@ void set_is_visible(const QWidget& w, bool force)
     }
     else
     {
+#if !defined(QT_NO_DEBUG_OUTPUT)
         eval_once(qDebug() << "check-visible: GetWindowRect failed");
+#endif
         visible = true;
     }
 }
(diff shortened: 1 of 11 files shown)


sthalik:
Since `eval_once` is only used for debug statements, I'd rather go for eliding the call entirely when `QT_NO_DEBUG_OUTPUT` is set:

```c++
#ifndef QT_NO_DEBUG_OUTPUT
...
#else
#   define eval_once(expr) void()
#endif
```

sthalik (changes requested):
Since `eval_once` is only used for debug statements, I'd rather go for eliding the call entirely when `QT_NO_DEBUG_OUTPUT` is set:

```c++
#ifndef QT_NO_DEBUG_OUTPUT
...
#else
#   define eval_once(expr) void()
#endif
```

sthalik:
Merged as 2f9b016c. Thanks.

PR #1867, opened 2024-06-08
Fix compatibility with Proton 9.0 and up

kzvezdarov:
## Problem
From Proton 9.0 and up, the directory containing library files and wine executable - previously called `dist`, appears to have been renamed to `files`. This breaks OpenTrack with newer Proton versions, because the path to the `wine` executable and proton library files is constructed as `:PROTON_PATH/dist/bin/wine`, so the executable is never found and tracking cannot be launched.

Proton 8.0 (note `dist`):
```bash
kiril@localhost ~/.l/s/S/s/common> ls Proton\ 8.0/
dist/       filelock.py*  LICENSE.OFL*  proton*                    proton_dist.tar*  toolmanifest.vdf*         version*
dist.lock*  LICENSE*      PATENTS.AV1*  proton_3.7_tracked_files*  __pycache__/      user_settings.sample.py*
```

Proton 9.0 (Beta) (`dist` is replaced by `files`):
```bash
kiril@localhost ~/.l/s/S/s/common> ls Proton\ 9.0\ \(Beta\)/
dist.lock*    files/    LICENSE.OFL*  proton*                    __pycache__/            steampipe_fixups.py*  user_settings.sample.py*
filelock.py*  LICENSE*  PATENTS.AV1*  proton_3.7_tracked_files*  steampipe_fixups.json*  toolmanifest.vdf*     version*
```

## Fix
To fix this while preserving backward compatibility, this commit changes `proton_path` to point to the `dist` dir directly, be it `files` for 9.0+ or `dist` for older versions. Template strings are adjusted accordingly, mostly removing the hardcoded `dist` part; as far as I can tell, the common parent dir for every path constructed from `proton_path` is the dist dir itself, so it seemed appropriate to fully replace the value of the original parameter, rather than introducing a new one.

Obtaining the dist. dir path for a specific Proton version is done by recursively iterating the version's directory, looking for the `wine` executable. The first match is used to discover the grandparent dir path, which is the desired `:PROTON_PATH/:DIST_DIR` path. This also means that only usable Proton directories - with a `wine` exectutable - will be listed in the dialog dropdown.

## Testing
I've built and tested this manually on OpenSuse Tumbleweed, with [Falcon BMS 4.37.4](https://www.falcon-bms.com/), [Nuclear Option](https://store.steampowered.com/app/2168680/Nuclear_Option/), and [DCS World Steam Edition](https://store.steampowered.com/app/223750/DCS_World_Steam_Edition/). Falcon BMS 4.37.4 is what led me to discover this issue, as Update 4 requires Proton 9.0+ for the alternative launcher to work, which broke my head tracking.

### Disclaimer
This is my first C++ in 15 years and first ever experience with Qt, so the code is likely far from idiomatic; happy to update with any suggestions.

### DIFF
--- a/proto-wine/proton.cpp
+++ b/proto-wine/proton.cpp
@@ -25,7 +25,7 @@ static const char* runtime_paths[] = {
 };
 
 
-std::tuple<QProcessEnvironment, QString, bool> make_steam_environ(const QString& proton_path, int appid)
+std::tuple<QProcessEnvironment, QString, bool> make_steam_environ(const QString& proton_dist_path, int appid)
 {
     using ret = std::tuple<QProcessEnvironment, QString, bool>;
     auto env = QProcessEnvironment::systemEnvironment();
@@ -35,7 +35,7 @@ std::tuple<QProcessEnvironment, QString, bool> make_steam_environ(const QString&
 
     auto expand = [&](QString x) {
                       x.replace("HOME", home);
-                      x.replace("PROTON_PATH", proton_path);
+                      x.replace("PROTON_DIST_PATH", proton_dist_path);
                       x.replace("RUNTIME_PATH", runtime_path);
                       return x;
                   };
@@ -58,14 +58,15 @@ std::tuple<QProcessEnvironment, QString, bool> make_steam_environ(const QString&
         error = QString("Couldn't find a Wineprefix for AppId %1").arg(appid);
 
     QString path = expand(
-        ":PROTON_PATH/dist/bin"
+        ":PROTON_DIST_PATH/bin"
+        ":PROTON_DIST_PATH/bin"
     );
     path += ':'; path += qgetenv("PATH");
     env.insert("PATH", path);
 
     QString library_path = expand(
-        ":PROTON_PATH/dist/lib"
-        ":PROTON_PATH/dist/lib64"
+        ":PROTON_DIST_PATH/lib"
+        ":PROTON_DIST_PATH/lib64"
         ":RUNTIME_PATH/pinned_libs_32"
         ":RUNTIME_PATH/pinned_libs_64"
         ":RUNTIME_PATH/i386/lib/i386-linux-gnu"
@@ -84,9 +85,9 @@ std::tuple<QProcessEnvironment, QString, bool> make_steam_environ(const QString&
     return ret(env, error, error.isEmpty());
 }
 
-QString proton_path(const QString& proton_path)
-{
-    return proton_path + "/dist/bin/wine";
-}
 
+QString proton_path(const QString& proton_dist_path)
+{
+    return proton_dist_path + "/bin/wine";
+}
 #endif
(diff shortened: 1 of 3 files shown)


proto-wine/proton.cpp
```diff
@@ -58,14 +58,15 @@ std::tuple<QProcessEnvironment, QString, bool> make_steam_environ(const QString&
         error = QString("Couldn't find a Wineprefix for AppId %1").arg(appid);
 
     QString path = expand(
-        ":PROTON_PATH/dist/bin"
+        ":PROTON_DIST_PATH/bin"
+        ":PROTON_DIST_PATH/bin"
```
sthalik, line 62: Copy-paste error?

sthalik (changes requested):

sthalik:
> This is my first C++ in 15 years and first ever experience with Qt, so the code is likely far from idiomatic; happy to update with any suggestions.
 
Actually the code quality is good. The only thing wrong is the copy-paste error (see the change request).

I'm going to need to test it for regressions in a VM with a Steam installation, it's going to take a few days to set it up. Maybe  @ljgdasfhk wishes to test it too.

proto-wine/proton.cpp
kzvezdarov: Yup, leftover from a previous iteration where I was just creating duplicate paths for setting `LD_LIBRARY`/`PATH`, fixed.

sthalik (approved):

sthalik:
Did you try it with Proton sandboxing turned on? Many users are having issues with it.

kzvezdarov:
> Did you try it with Proton sandboxing turned on? Many users are having issues with it.

I haven't but I can try - what kind of sandboxing are you referring to in this case, using AppArmor/Firejail?

ljgdasfhk:
> Maybe @ljgdasfhk wishes to test it too.

I'm sorry, friends, my main operating system now is Windows, and not Linux, as before. If I notice something not working correctly, I will report it immediately.

sthalik:
> I haven't but I can try - what kind of sandboxing are you referring to in this case, using AppArmor/Firejail?

The one that's possible to enable in Steam's per-game setting. Then launch using "Play" in Steam as normal.

kzvezdarov:
> > I haven't but I can try - what kind of sandboxing are you referring to in this case, using AppArmor/Firejail?
> 
> The one that's possible to enable in Steam's per-game setting. Then launch using "Play" in Steam as normal.

Oh do you mean the compatibility tab, e.g.:
Screenshot_20240609_222239

If that's the case, then I've tested that for this branch with both Proton 8.0 (the old version which still works on `master`) and 9.0+ (the new version, which is broken due to the `dist/ -> files/` directory rename), specifically with DCS World and Nuclear Option (I play and launch those through Steam, with the compatibility tool set to the specific version that works for the game).

sthalik:
Merged, thanks.

PR #1870, opened 2024-06-17
tracker/nn: Replace models with re-exports which aren't broken with onnxruntime 1.18

DaWelter:
Hi @sthalik ,

Regardless of downgrading onnxruntime, why not?

There is hardly a downside. The model outputs should be essentially the same as before. Only the predicted head position will be offset to further back by a fixed translation.


sthalik:
In this commit you're removing models from the tree rather than replace them with updated versions. Is this what you want to do?

PR #1872, opened 2024-06-20
filter/accelahamilton: rewrite in double precision

DaWelter:
Self explanatory.

Reason: strange stuttering motion after a while that I observed with this filter, but not with Accela.
Also, the Accela and Hamilton filter are written in double precision. So it's probably better to not take chances for bad result due to limited precision of float.

Maybe it's confirmation bias but I think the output is nice and smooth now.

Thanks @tombrazier and @GO63-samara for the math library :-D

### DIFF
--- a/filter-accela-hamilton/accela_hamilton.cpp
+++ b/filter-accela-hamilton/accela_hamilton.cpp
@@ -6,14 +6,13 @@
  * copyright notice and this permission notice appear in all copies.
  */
 #include "accela_hamilton.h"
-#include "compat/math.hpp"
 #include "api/plugin-api.hpp"
 #include "opentrack/defs.hpp"
 
 #include <algorithm>
-#include <QDebug>
-#include <QMutexLocker>
 
+#include "compat/math.hpp"
+#include "compat/hamilton-tools.h"
 #include "compat/math-imports.hpp"
 #include "compat/time.hpp"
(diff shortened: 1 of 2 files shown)


sthalik (approved):

GO63-samara:
You have done a hard job, but I have some doubts.

In my opinion, `float` precision is quite sufficient.
The idea of ​​abandoning `tQuat` in favor of `QQuaternion` was quite reasonable, since `QQuaternion` is already used in other Opentrack modules.

What do you mean by `strange stuttering motion`?
Perhaps the strange stuttering motion is associated with an incorrect transition from Euler angles to quaternions and back or something similar, and not at all with the accuracy of the numbers ?

But maybe you're right.

DaWelter:
@GO63-samara , I mean the output felt "quantized", i.e. piecewise constant output.

@sthalik  was so kind to make me a build with his settings and compared to his previous build there was a clear difference. Much smoother output.

As strange as it seems. There are other explanations than pure numerical precision. Choice of "normalization" constants, clipping thresholds and all the other devilish implementation details.

Anyhow I'm quite happy with your tQuat and tVector. They also get the job done. :-)

GO63-samara:
Okay, now I understand.
I have no more doubts.
I support this.

PR #1888, opened 2024-07-07
proto-wine: Add search paths for debian-installation

MordiJC:
Steam installer on Ubuntu places its files in $HOME/.steam/debian-installation/
This patch fixes error: "Couldn't find steam runtime"


sthalik:
Thanks!

Issue #1923, opened 2024-09-04
RFC: Linux - Extend Proton support beyond Steam to keep compatability with launchers (like Lutris) that use UMU


Priton-CE:
### The Problem
Lutris is starting to adapt [UMU](https://github.com/Open-Wine-Components/umu-launcher) which utilizes Proton for execution.

This is causing problems with titles run through Lutris with the UMU backend, as now instead of normal wine **they are using Proton for execution which Opentrack only supports in the context of Steam**.

(See #1869 as an example and for the current workaround which utilizes the current proton launch method.)

### The current state of Wine/Proton Support
Currently the proto-wine module supports two ways of launching the wrapper program inside of a wine/proton prefix:

1. Take a path to the wine executable and a path to the prefix and launch the wrapper script using these parameters. _(Optionally Opentrack also looks through the Lutris Runner Directory to suggest possible wine paths)_
2. Take a proton path and a steam_id to launch the wrapper using the proton path (and proton environment variables) inside the wine prefix corresponding to the steam_id. _(To find a proton path Opentrack looks through the `.../steamapps/common` folder as it houses the official Proton builds)_

What is to note here is that

- When Proton is used by other Game Launchers (like Lutris which it can now do thanks to UMU) Opentracks Wrapper Program will not be able to interface with them as its limited to run inside steams prefixes.
- Only official Proton Versions can be used as 3rd party Proton will not be installed to the `.../steam/steamapps/common` folder. This limits the use of 3rd party Proton versions like the ones made by Glorious Eggroll within Steam and also severely limiting possible support for UMU launched titles, as it prefers using GEs proton builds.

I dont know why the Proton feature is limited to Steam in this way. I assume its to make using it easier but it has the sideeffect that, now that Proton can be used outside of Steam, Opentrack does not work with Proton run titles outside of Steam.

### What needs to be done
To support Proton games launched with Lutris one would need to allow the proton method to use proton versions and wine prefixes other than the ones created by steam (decoupling the proton feature from the steam echosystem but not destroying support).
That means:

1. The Dropdown to select Proton Versions would need to also contain the results from `.../steam/compatibilitytools.d/`
2. Instead of just taking a Steam Application ID the user would be able to select either the steam_id option or a custom path to a wine prefix

(This assumes that the current proton launch method can launch the wrapper in a way that allows it to communicate with UMU launched titles, which I dont see any reason why it should not that, especially after reports of a working workaround in the above issue)

### Conclusion
[...]

Issue #1879, opened 2024-06-25
proto/wine can be done easier


sthalik:
Actually rather than doing the whole dance with creating a shared memory mapping on the Wine side in `proto/wine`, we can create a `c:/opentrack-memory-mapping.dat` of the right size and map it from both native ELF and `.dll` with stable ABI.

sthalik:
Once this is done, all the stupid nonsense with Proton `winegcc` ABI will be a thing of the past.

Priton-CE:
Would this even need to be implemented on an ABI level?
What arguments are there against just serializing the data into a format like JSON? The only argument I can really think of here is that it adds a bit of overhead having to serialize and deserialize, but it would remove the need for a consistent ABI.

Maybe I also don't understand quite right but how would the ABI be consistent across the Opentrack binary and the wrapper binary. I thought winegcc would be the thing that ensures the ABI is identical between both binaries.

Tho doing a bit of a rework on the proto-wine module would be appreciated. I found it quite hard to understand what components do what exactly.

Priton-CE:
@sthalik since this topic came up in the "UMU Proton using opentrack-git" issue:
The idea here is to

1. create a file in the prefix
2. using a registry key map this file into memory | [microsoft docs](https://learn.microsoft.com/en-us/windows/win32/memory/file-mapping)
3. have the trackir protocol use this memory location instead

correct?

sthalik:
(1) is right

2\. Then use the registry to point to `NPClient{,64}.dll` (https://github.com/opentrack/opentrack/blob/master/contrib/npclient/npclient.c) which is modified for this purprose.

Instead of creating the *named* memory mapping (https://github.com/opentrack/opentrack/blob/master/compat/shm.h) as before, it maps a *file* in the same manner using this struct layout (https://github.com/opentrack/opentrack/blob/master/freetrackclient/fttypes.h#L59-L68).

There's this simple test app that works for the non-"trackir enhanced" testing at least: https://github.com/opentrack/opentrack/tree/master/contrib/freetracktest

(3) is right, NPClient sends it to the game.

There's also `proto/ft` which does the same things as `proto/wine` using the same shared memory mapping format.

I should manage to finish it eventually, probably even faster than fixing the Windows build issues (*sigh*)

PR #1932, opened 2024-09-21
Extend Proton Support in the proto-wine module

Priton-CE:
RFC: #1923 

### Problem
The proto-wine module has Proton support but it only allows Proton to be used within Steam. With the introduction of UMU and the shift of GloriousEggroll towards Proton the current feature support may no longer be adequate.

### Solution provided by this PR
The UI has been changed to allow the user to specify a Proton Prefix or keep on utilizing the already existing Steam Application ID mechanism.
Elements that are not in use or relevant to the current functionality are also now being disabled.

Screenshot_20240921_011346

Additionally the `make_steam_environ(proton_path, appid)` function has been split up into `make_steam_environ(proton_path)` and `make_wineprefix(appid)` to continue using the steam environment independently from the appid.

The `wine::initialize()` function has been refactored to make use of the above changes and add comments explaining every step.

### Change Summary
- Redesigned the wine UI to allow for specifying custom Proton Prefixes and improve UX
  - The user is promted to select either the Wine or Proton environment (disabling the UI components of the other)
  - If Proton is chosen the user further needs to specify is Proton is used within Steam or externally (likely through UMU). Tooltips are present to explain what each option does. (This also features the disabling of UI components.)
- Refactored protocol code to allow for the use of Proton outside of Steam
- Fixed an issue where Proton Versions may appear multiple times
- Fixed an issue where some 3rd party Proton Versions were not getting detected

### DIFF
--- a/proto-wine/ftnoir_protocol_wine_dialog.cpp
+++ b/proto-wine/ftnoir_protocol_wine_dialog.cpp
@@ -5,8 +5,10 @@
 #include <QDirIterator>
 #include <qdebug.h>
 #include <qdir.h>
+#include <qradiobutton.h>
 
 #include "api/plugin-api.hpp"
+#include "options/tie.hpp"
 
 /*
  * 0: path to the directory with wine versions
@@ -28,7 +30,7 @@ FTControls::FTControls()
 {
     ui.setupUi(this);
 
-    //populate wine select
+    // populate wine select
     ui.wine_path_combo->addItem("System Wine", QVariant{"WINE"});
     for (const char** path : wine_paths) {
         QDir dir(QDir::homePath() + path[0]);
@@ -44,7 +46,7 @@ FTControls::FTControls()
     }
     ui.wine_path_combo->addItem("Custom path to Wine executable", QVariant{"CUSTOM"});
 
-    //populate proton select
+    // populate proton select
     for (const char* path : proton_paths) {
         QDir dir(QDir::homePath() + path);
         dir.setFilter(QDir::Dirs);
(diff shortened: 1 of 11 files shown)


proto-wine/ftnoir_protocol_wine_dialog.cpp
```diff
@@ -103,6 +124,63 @@ void FTControls::onWinePathComboUpdated(QString selection) {
...
 
+void FTControls::onRadioButtonsChanged() {
+    if (ui.variant_wine->isChecked()) {
+        // wine settings selected
+
+        // enable wine settings
+        ui.wine_path_combo->setEnabled(true);
+        ui.wineprefix->setEnabled(true);
+        ui.browse_wine_prefix_button->setEnabled(true);
+        if (ui.wine_path_combo->currentText() == "Custom path to Wine executable") {
```
sthalik, line 135: This is bad, absolutely don't do that. Use [QComboBox::itemData](https://doc.qt.io/qt-6/qcombobox.html#itemData) (just stick the actual path in there).

proto-wine/ftnoir_protocol_wine_dialog.cpp
```diff
@@ -5,8 +5,10 @@
 #include <QDirIterator>
 #include <qdebug.h>
 #include <qdir.h>
+#include <qradiobutton.h>
```
sthalik, line 8: It's idiomatic in Qt to include `<QRadioButton>` even though your IDE suggests otherwise. The same issue is present in several other files.

sthalik:
It's a workable solution until after #1879 is resolved. Then there shouldn't be any more sandboxing issues.

Too bad that 80% of the code changes pertain solely to the UI. But there doesn't seem to be an obvious solution of doing it in a much shorter way, or avoid doing the UI work altogether.

But I won't be able to test this on a Linux box for a while. At least don't introduce regressions for non-Proton Wine, okay?

The indent style is pretty inconsistent by now but you haven't started it so it's not really your fault.

sthalik (changes requested):

sthalik:
Oh, and -- see if the [Qt tab order](https://doc.qt.io/qt-6/designer-tab-order.html) is correct, it tends to break itself.

Priton-CE:
@sthalik I will see to implementing that change and checking the tab order.

Regarding the indent style I dont see much of a different way of doing this, at least if we want to keep the Proton settings confined to one Proton Radio Button.
I suppose instead of doing
```
- Wine (wine version, [custom path], wine prefix)
- Proton (proton version)
  - SteamPlay (app id)
  - UMU (proton prefix)
```
one could also just go for
```
- Wine (wine version, [custom path], wine prefix)
- Proton SteamPlay (proton version, app id)
- Proton UMU (proton version, proton prefix)
```

That would clean up the indent style a bit but it would duplicate the proton version combo box.
But at least it would enforce the style of: "Give me a version and give me a place to run" across the board instead of "give me a version and then tell me how to run and where to run" specifically for the Proton option.

Regarding regressions: I made sure to test the wine option as well but I will make sure to retest everything after implementing your changes.
The control flow of the wine option should be pretty much identical so I don't think there would be much of a risk of me missing things there.

proto-wine/ftnoir_protocol_wine_dialog.cpp
sthalik: In the future we might want to append some sort of indicator to the display text as opposed to the literal name stored in the .ini.

sthalik:
@Priton-CE I was only referring to non-functional style issues such as where the braces are placed. For the purpose of this PR just leave it as-is, it's easier to fix all of them in one go.

Actually your original way of aligning the radio checkboxes is more intuitive to the end-user than the flattened style.

proto-wine/ftnoir_protocol_wine_dialog.cpp
Priton-CE: Changes have been implemented.
Closing this request tho I still gotta do the Wine and Proton test. Will put a message once I completed those.

Priton-CE:
@sthalik
I have implemented the requested changes (switching from the combo text to combo data) and have fixed the tab order which was indeed messed up.
I also completed regression tests with Wine and Proton. Both continued to work. (I can't test Proton via SteamPlay since I don't have any headtracking Steam games but there shouldn't be any functional differences other than the path generation, which I only moved into a separate function.)

PR #1943, opened 2024-10-04
macOS Build fixes, changes to port-wine, changes to the X-Plane-plugin

matatata:
Hello Stanislav,

I have made a couple of tweaks that I'd like you to review and maybe merge.

Most notably:
- I updated the x-plane plugin so that it plays nice with a popular plugin called X-Camera. I also added menu-items to control and see the plugins status.
- Then I finally was able to build the proto-wine module on macOS and it does work with TrackIR enabled games on macOS ! However I did not like the fact that once the module is built with wine support you always must wine installed and you cannot just the native shared memory method. For that I added an option that so you can choose to not use wine.

 I plan to test things that could affect the linux build but I haven't had a chance yet. I also don't have X-Plane for linux but I'm confident I did not break anything.

Thank you in advance & best regards, Matteo


macosx/make-app-bundle.sh
```diff
@@ -42,6 +42,15 @@ cp -r "$install/Plugins" "$install/$APPNAME.app/Contents/MacOS/"
 macdeployqt "$install/$APPNAME.app" -libpath="$install/Library"
 #sh "$dir/install-fail-tool" "$install/$APPNAME.app/Contents/Frameworks"
 
+
```
matatata, line 47: Actually I think it's weird that this library even is a dependency. It shows up in the neuralnet-tracker.dylib. But it also shows up in the opentrack-wrapper.exe.so which was compiled by wineg++  with an absolute path /opt/local/lib/libundwind.1.dylib . That can't be right. No one will have that. I read things that it will be picked up and linked more or less by accident https://trac.macports.org/ticket/66250. not sure if it has something to do with that. In order to make opentrack-wrapper.exe.so work for others  I had to replace that reference via install_name_tool

matatata: Actually MacPorts fixed it recently :-) https://trac.macports.org/ticket/71052 . No more absolute references to libunwind in the build :-) I'll update the commit/PR

Priton-CE:
I dont really get that wine change.

Its a wine module, why launch without wine?

Priton-CE:
Also calling it Wine/X-Plane makes no sense

This is a compatibility layer for wine. There is no dependency on X-Plane. This would be like me calling it Wine/StarCitizen.

Priton-CE:
My main concern here is that a user comes along and sees the following:

"I have this wine module! Cool! Ok but why can I launch without wine?"

The naming just isnt clear or intuitive. If there is an option to launch without wine the naming should be changed.

matatata:
Hi @Priton-CE  , so I have reverted my changes to the proto-wine and split the porto-wine module into two modules:
proto-wine basically stays the same it was, but I moved everything about that X-Plane business into a new module proto-native-shm. So both modules now are less complex and Linux and macOS users can now use the wine module for and only for Wine and we macOS guys and Linux users can use the new module with X-Plane by using opentrack's x-plane plugin where the plugin reads the motion data directly from shared memory. It really works well with native X-Plane on Linux by the way. In theory other native linux software could use the same module given that someone writes a plugin or some code that accesses that shared memory as well. Therefore I did not name the module proto-xplane, because although currently I guess there's only one client i.e. the X-Plane-Plugin.

matatata:
Hi, I'll close this PR for now as I'm currently making a lot of modifications regarding the WINE integration: Main reason is that compiling opentrack-wrapper-wine.exe.so on unix with clang is no longer supported. On macOS seems that this winelib does not work anymore since wine 10.x and it never worked well with CrossOver. I now can't even compile the wrapper on macOS anymore [1]. As Alexandre Julliard suggested [2] one could use a different approach called UNIX-Call interface instead. I'm currently working on modified versions of freetrackclient.dll and npclient.dll that will now call into an accompanying native posix shared library that can access the native shared memory such as the WineSHM structure. Those dlls are also known as "builtin" wine dlls.
As a consequence there's no need to copy the data in a loop anymore and it more or less means that opentrack hasn't really much to do expect maybe put the location of fake-dlls into the registry but then that could be done with a simple .exe that the user needs to run themselves inside their WINEPREFIX or Bottle. Also those builtin dlls need to be copied into the wine installation <install>/lib/wine/x86_64-windows and <install>/lib/wine/x86_64-unix .

I saw that there were discussion considering to write a file instead of shared memory so that it could be read from anywhere. I saw this only after I started my work so for now I'll try to polish what I already have. I'll then probably open a new Pull Request.

References:
[1] https://trac.macports.org/ticket/72792
[2] https://bugs.winehq.org/show_bug.cgi?id=58580

PR #1954, opened 2024-11-02
tracker/nn: Update models

DaWelter:
Hi @sthalik 

I think it is time to update the neuralnet models.  :-)

For one I removed `head-pose.onnx`. I think the newer models are better. Personally I don't use it any more.

Then I added `head-pose-0.3-big-quantized.onnx`. It is the same architecture as `head-pose-0.2`. Performance in terms of accuracy should be very similar. However this one was "quantized" after training, i.e. the convolution ops were converted so they use integer computations. On my machine this reduces inference time to 70% of the float32 model with negligible accuracy loss.

The mobilenet variant suffered quite a big hit which is why only the big model is updated.

I left `head-pose-0.2` in place because I don't know if the quantized model has universally better performance on all hardware. To get some user experiences would be interesting.

(New model trained with latest https://github.com/opentrack/neuralnet-tracker-traincode commit 9891d60)

### DIFF
--- a/tracker-neuralnet/CMakeLists.txt
+++ b/tracker-neuralnet/CMakeLists.txt
@@ -33,9 +33,9 @@ if(OpenCV_FOUND AND ONNXRuntime_FOUND AND OpenMP_FOUND)
 
     install(
         FILES "models/head-localizer.onnx" 
-              "models/head-pose.onnx"
               "models/head-pose-0.2-big.onnx"
               "models/head-pose-0.2-small.onnx"
+              "models/head-pose-0.3-big-quantized.onnx"
         DESTINATION "${opentrack-libexec}/models"
         PERMISSIONS ${opentrack-perms-file}
         )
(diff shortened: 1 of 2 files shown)


sthalik:
Merged, thanks!

MysticDaedra:
I presume this is running solely on CPU? My Ryzen 7 5700X @ 4.8GHz gets 14-15ms inference time with the new quantized version, and 7ms inference time on both 0.2 versions. I'd say for some folks (me!), the 0.2 versions are both adequate and considerably faster.

DaWelter:
Hi, Thanks for the feedback. Interesting. Will keep it in mind and keep the float 32 models.

MysticDaedra:
Further use seems to have clarified a few things. I definitely get better
facial/head recognition with the new quantized big model, and performance
is actually pretty decent, got it down to ~10ms average, but I had to
increase the thread count to 4 and decrease the resolution slightly. I
actually think it's a fair trade-off, but yeah I'd recommend keeping the
older models around for a bit longer for better compatibility with
slower/older hardware.

PR #1995, opened 2025-03-04
tracker/nn: Add more fps options

DaWelter:
This PR is an attempt to add more FPS options to the neuralnet tracker.

@sthalik , or anyone, is this all I need to change?
I check that the fps is passed to your `camera::start`. That's okay.
But on linux my (cheap) webcam was still fixed at 30 fps and the trackhat cam at 60 fps. Need to check on windows.


sthalik:
You also need to add it to the UI.

DaWelter:
UI actually look ok. The dropdown menu is generated in `make_fps_combobox()`.
And my camera supports only 30fps so I can't test it. :-/

DaWelter:
Thanks. :+1:

PR #2002, opened 2025-03-22
Give evdev arbitrary bustype vendor product and version ids

Spaceslug:
This fixes Euro Truck Simulator 2 on linux not detecting the libevdev joystick receiver.


sthalik:
Merged, thanks!

PR #2003, opened 2025-03-23
tracker/nn: Fix roll output

DaWelter:
Hi guys,

I think there might have been a little bug since the beginning in the NN tracker ...

Roll is now (hopefully) computed from the correct matrix. It turns out it was using values from row 2 instead of 1. Makes more sense to measure roll based on the y-component of local axes. Must have slipped in the matrix row somehow. Reference to wikipedia is added.

Only noticed now where I experimented with centering methods and a model which can handle large yaws. Somehow the previous roll computation was working okay for smallish yaw. Near 90 deg it outputted garbage.


sthalik:
Merged, thanks!

PR #2015, opened 2025-04-13
Deferred neck transformation so it's more akin to DCS's implementation.

cecil1527:
See discussion https://github.com/opentrack/opentrack/discussions/2011


sthalik:
Not sure if it's actually good to do a hard clamp to zero below 90 deg. It may as well be steep, just not like that.

cecil1527:
Yeah, I suppose you could customize it however you want. I essentially wanted the DCS neck model in IL-2, so that's why I made those changes. 

I think it feels really natural. I don't want my head translating as I look around the front of the cockpit at gauges and displays. I only want the automatic neck translation to kick in once I go to look over my shoulder, behind me. But I only play flight sims, so I don't know what others might want.

PR #2017, opened 2025-04-21
Finds the closest body from the center of sensor viewport if any

searching46dof:
Implements face trascking priority closest to center of viewport


searching46dof:
Implements face trascking priority closest to center of viewport

PR #2026, opened 2025-05-06
Update CMake workflow for Ubuntu 22.04 and ONNXRuntime integration

TLCFEM:
I saw the current CI pipeline does not use a consistent Qt installation and does not pack the binary, I try to improve the workflow in this PR.

### DIFF
--- a/.github/workflows/linux-package.yml
+++ b/.github/workflows/linux-package.yml
@@ -2,8 +2,7 @@ name: Linuax Package
 on:
   push:
     tags:
-      - 'opentrack-[0-9]+.[0-9]+.[0-9]+'
-      - 'opentrack-[0-9]+.[0-9]+.[0-9]+-[a-zA-z0-9]+'
+      - 'opentrack-*'
   workflow_dispatch:
 env:
   build_type: Release
@@ -63,7 +62,7 @@ jobs:
         echo '"$CURRENT_PATH/bin/opentrack" "$@"' >> opentrack
         chmod +x opentrack
 
-        if [[ ${{github.ref_name}} =~ ^opentrack-\d+\.\d+\.\d+$ ]]; then
+        if [[ ${{github.ref_name}} =~ ^opentrack.*$ ]]; then
           file_name=${{github.ref_name}}-linux.tar.gz
         else
           file_name=opentrack-dev-linux.tar.gz


.github/workflows/cmake.yml
```diff
@@ -18,12 +19,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
```
TLCFEM, line 21: `ubuntu-latest` uses 24.04, which may be too new.
Using an older distro here so that glibc won't be too new for most users.

TLCFEM:
Based on this PR, I can use [linuxdeployqt](https://github.com/probonopd/linuxdeployqt) to generate a self-containing bundle, as can be seen in the following workflow.

https://github.com/TLCFEM/opentrack/actions/runs/14850230733

Not sure if you are interested in fully automated CICD pipelines, though.

cmake/opentrack-install.cmake
```diff
@@ -50,9 +50,12 @@ endif()
 if (APPLE)
     set(OSX_POST_INSTALL_DIR "/../thirdparty")
 endif()
-otr_install_exec("${opentrack-libexec}${OSX_POST_INSTALL_DIR}" FILES "${CMAKE_SOURCE_DIR}/bin/freetrackclient.dll")
-otr_install_exec("${opentrack-libexec}${OSX_POST_INSTALL_DIR}" FILES "${CMAKE_SOURCE_DIR}/bin/freetrackclient64.dll")
-otr_install_exec("${opentrack-libexec}${OSX_POST_INSTALL_DIR}" FILES
-    "${CMAKE_SOURCE_DIR}/bin/NPClient.dll"
-    "${CMAKE_SOURCE_DIR}/bin/NPClient64.dll"
-    "${CMAKE_SOURCE_DIR}/bin/TrackIR.exe")
```
sthalik, line 58: Don't do that. It's for Wine.

.github/workflows/cmake.yml
sthalik: This file is only for compile-time and (hopefully at some point) runtime tests.

Make a new Linux-only build script based on `cmake.yml`. Name it something like "ubuntu-package". Make it run only on tags and only on your preferred old-but-not-EOL LTS branch.

tracker-neuralnet/CMakeLists.txt
```diff
@@ -1,8 +1,4 @@
 include(opentrack-opencv)
-set(host-spec "${CMAKE_SYSTEM_NAME} ${CMAKE_SYSTEM_PROCESSOR} ${CMAKE_SIZEOF_VOID_P}")
-if(host-spec MATCHES "^Linux i[3-6]86 4$")
-    return()
-endif()
```
sthalik, line 5: Don't, this as it actually fixes an x86 build failure.

.github/workflows/cmake.yml
```diff
@@ -43,18 +46,19 @@ jobs:
...
+      uses: jurplel/install-qt-action@v4
       with:
-        archives: qtbase qtimageformats qtgamepad qttools qtserialport qtmultimedia
-      if: matrix.os != 'ubuntu-latest'
+        version: 5.15
+        archives: icu qtbase qtimageformats qtgamepad qttools qtserialport qtmultimedia
 
     - name: Configure
-      run: ${{matrix.cmake}} -GNinja -S ${{github.workspace}}/ -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DQt5_DIR=${{env.Qt5_DIR}} -DQt5Gui_DIR=${{env.Qt5_DIR}}/lib/cmake/Qt5Gui
+      run: ${{matrix.cmake}} -GNinja -S ${{github.workspace}}/ -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_PREFIX_PATH=${{env.Qt5_DIR}} -DONNXRuntime_DIR=${{matrix.onnx_prefix}}-${{matrix.onnx_version}}
```
sthalik, line 61: Don't' touch `Qt5_DIR`. It's not even correct to be using `CMAKE_PREFIX_PATH` for just one dependency out of several. The bit about `Qt5Gui_DIR` maybe could even be removed.

I suppose using the official package is better than relying on some arbitrary PPA.

.github/workflows/cmake.yml
```diff
@@ -43,18 +46,19 @@ jobs:
...
+      if: contains(matrix.os, 'ubuntu')
 
     - name: Install Qt
-      uses: jurplel/install-qt-action@v3
+      uses: jurplel/install-qt-action@v4
       with:
-        archives: qtbase qtimageformats qtgamepad qttools qtserialport qtmultimedia
-      if: matrix.os != 'ubuntu-latest'
+        version: 5.15
+        archives: icu qtbase qtimageformats qtgamepad qttools qtserialport qtmultimedia
```
sthalik, line 58: You should't need qtmultimedia or qtgamepad.

sthalik (changes requested):

.github/workflows/linux-package.yml
```diff
@@ -0,0 +1,77 @@
...
+      with:
+        version: 5.15
+        archives: icu qtbase qttools qtserialport
+
+    - name: Configure and Build
+      run: |
+        cmake -S . -B build \
+          -DCMAKE_BUILD_TYPE=${{env.build_type}} \
+          -DONNXRuntime_DIR=${{env.onnx_prefix}}-${{env.onnx_version}} \
+          -DQt5_DIR=${{env.Qt5_DIR}}
```
TLCFEM, line 43: The official documentation actually recommends using `CMAKE_PREFIX_PATH`, over `Qt5_DIR`, see [here](https://doc.qt.io/qt-5/cmake-get-started.html).

.github/workflows/linux-package.yml
```diff
@@ -0,0 +1,77 @@
+name: Linuax Package
+on:
+  push:
+    tags:
+      - 'opentrack-\d+\.\d+\.\d+'
```
TLCFEM, line 5: Maybe it is okay?

TLCFEM:
Screenshot from 2025-05-06 10-58-53

The package is self-containing, and it appears to work fine.

.github/workflows/linux-package.yml
sthalik: There are sometimes suffixes like `-rc1` or `-beta`. Maybe try:
`opentrack-\d+\.\d\S+`

.github/workflows/linux-package.yml
sthalik: It's probably good enough. But at least do this for correctness' sake.

```cmake
list(APPEND CMAKE_PREFIX_PATH ${{env.Qt5_DIR}})
```

.github/workflows/linux-package.yml
sthalik: These are Git-style globs. So something like: `opentrack-[0-9]*.[0-9]*`.

.github/workflows/linux-package.yml
TLCFEM: It's not a CMake file though.

.github/workflows/linux-package.yml
```diff
@@ -0,0 +1,78 @@
+name: Linuax Package
+on:
+  push:
+    tags:
+      - 'opentrack-[0-9]+.[0-9]+.[0-9]+'
+      - 'opentrack-[0-9]+.[0-9]+.[0-9]+-[a-zA-z0-9]+'
```
TLCFEM, line 6: For `-rc` tags.

sthalik: These are globs, you can't use `+` with its regex meaning.

Hell, just make it `opentrack-*` and leave it at that.

TLCFEM: Don't worry, it is fully compatible with the doc. https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

.github/workflows/linux-package.yml
```diff
@@ -0,0 +1,77 @@
+name: Linuax Package
+on:
+  push:
+    tags:
+      - 'opentrack-*'
```
TLCFEM, line 5: as required

sthalik:
Merged, thanks!

PR #2062, opened 2025-08-12
tracker/neuralnet: revert construction of Ort::Env mostly

DaWelter:
Hi @sthalik , 

like so?

As far as I can tell, the change to `Ort::Env` wasn't needed. This PR works for me, too. Ort::Env uses `Ort::Global<void>::api_` internally, so it doesn't need special treatment over the other c++ classes which do the same.

### DIFF
--- a/tracker-neuralnet/CMakeLists.txt
+++ b/tracker-neuralnet/CMakeLists.txt
@@ -3,9 +3,8 @@ set(host-spec "${CMAKE_SYSTEM_NAME} ${CMAKE_SYSTEM_PROCESSOR} ${CMAKE_SIZEOF_VOI
 if(host-spec MATCHES "^Linux i[3-6]86 4$")
     return()
 endif()
-if(NOT DEFINED opentrack-use-onnxruntime-avx-dispatch)
-    set(opentrack-use-onnxruntime-avx-dispatch 0)
-endif()
+
+set(opentrack-use-onnxruntime-avx-dispatch False CACHE BOOL "Enable dynamic loading and dispatching to avx-enabled onnxruntime")
 
 find_package(OpenCV QUIET)
 find_package(OpenMP QUIET) # Used to control number of onnx threads.
--- a/tracker-neuralnet/env.cpp
+++ b/tracker-neuralnet/env.cpp
@@ -45,7 +45,7 @@ bool is_avx_supported()
 }
 #endif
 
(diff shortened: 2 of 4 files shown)


tracker-neuralnet/CMakeLists.txt
```diff
@@ -3,9 +3,8 @@ set(host-spec "${CMAKE_SYSTEM_NAME} ${CMAKE_SYSTEM_PROCESSOR} ${CMAKE_SIZEOF_VOI
 if(host-spec MATCHES "^Linux i[3-6]86 4$")
     return()
 endif()
-if(NOT DEFINED opentrack-use-onnxruntime-avx-dispatch)
-    set(opentrack-use-onnxruntime-avx-dispatch 0)
-endif()
+
+set(opentrack-use-onnxruntime-avx-dispatch False CACHE BOOL "Enable dynamic loading and dispatching to avx-enabled onnxruntime")
```
sthalik, line 7: It was fine without the cache variable. It's meant to be an implementation detail for someone using the sdk-paths mechanism. It won't even install properly without cmake `install()` for the dll in sdk paths.

This isn't even a properly named cache variable. They use grouping, i.e. `opentrack_` and tend to form some kind of hierarchy, e.g.

`opentrack_tracker-neuralnet_use-avx-dispatch`

tracker-neuralnet/env.cpp
```diff
@@ -54,48 +54,30 @@ Ort::Env NeuralNetTracker::make_ort_env()
...
-        std::abort();
-    }
-    return Ort::Env{env};
-#else
-    return Ort::Env {
-        OrtLoggingLevel::ORT_LOGGING_LEVEL_ERROR,
-        "tracker-neuralnet"
-    };
+    Ort::Global<void>::api_ = ort_api; // see ORT_API_MANUAL_INIT in the onnx c++ header.
 #endif
```
sthalik, line 81: Great to have it simplified, good catch.

sthalik:
FWIW, I've committed the non-cmake changes under your name as 046cfb1c41fa.

tracker-neuralnet/CMakeLists.txt
DaWelter: Yeah, I thought you might say that. Leave it as is then.

DaWelter:
All right, this PR can be closed then.

PR #2092, opened 2025-10-20
Fix build with Qt 6.10

### LINKED ISSUE #2091: Arch: install fail qpa/qplatformnativeinterface.h: No such file or directory
Image

failing cause of qpa file/folder missing?

engdyn:
This fixes #2091 
> Usage of a private Qt module Foo now requires a call to find_package(Qt6 COMPONENTS FooPrivate) to make the Qt6::FooPrivate target available.


sthalik:
Hey,

```patch
diff --git a/cmake/opentrack-qt.cmake b/cmake/opentrack-qt.cmake
index 39aa6361..38d50171 100644
--- a/cmake/opentrack-qt.cmake
+++ b/cmake/opentrack-qt.cmake
@@ -4,7 +4,13 @@ set(qt-optional-components SerialPort)
 set(qt-imported-targets Qt6::Core Qt6::Gui Qt6::Network Qt6::Widgets)
 if(APPLE)
     list(APPEND qt-required-components Multimedia)
-    list(APPEND qt-imported-targets Multimedia)
+    list(APPEND qt-imported-targets Qt6::Multimedia)
+endif()
+
+if(NOT APPLE AND NOT WINDOWS)
+    list(APPEND qt-required-components GuiPrivate)
+    list(APPEND qt-imported-targets Qt6::GuiPrivate)
+    set(QT_NO_PRIVATE_MODULE_WARNING 1)
 endif()
 
 find_package(Qt6 REQUIRED COMPONENTS ${qt-required-components} QUIET)
diff --git a/qxt-mini/CMakeLists.txt b/qxt-mini/CMakeLists.txt
index c9b079ad..40ca0820 100644
--- a/qxt-mini/CMakeLists.txt
+++ b/qxt-mini/CMakeLists.txt
@@ -2,6 +2,9 @@ if(UNIX OR APPLE)
     set(self opentrack-qxt-mini)
     include_directories(SYSTEM ${Qt6Gui_PRIVATE_INCLUDE_DIRS})
     otr_module(qxt-mini NO-COMPAT BIN)
+    if(NOT WINDOWS AND NOT APPLE)
+        target_link_libraries(${self} PRIVATE Qt6::GuiPrivate)
+    endif()
     if(APPLE)
         find_library(CoreFoundation CoreFoundation)
         find_library(Carbon Carbon)
```

I was able to reproduce this issue on my Arch box, but would rather not assume that Linux is the only X11 platform. I'd like to commit this changed patch under your name instead. Does this work for you?

engdyn:
Hey, the problem is that Ubuntu has ancient Qt packages that don't have the Qt6::GuiPrivate target. You'd need to add a version check or else it breaks

sthalik:
I don't have old Ubuntu available, but maybe if we replace the condition with:

```
if(TARGET Qt6::GuiPrivate AND NOT APPLE AND NOT WINDOWS)
```

engdyn:
```
if(TARGET Qt6::GuiPrivate AND NOT APPLE AND NOT WINDOWS)
```
Doesn't work so I replaced it with:
```
if(NOT APPLE AND NOT WINDOWS AND QT_VERSION VERSION_GREATER_EQUAL 6.10.0)
```
You also need to add `find_package()` or else the Qt6::GuiPrivate target doesn't exist.

sthalik:
Merged. Thanks for bearing with me.

PR #2096, opened 2025-11-06
Change libevdev id vendor and id product to arbitrary unique values. Fixes Euro Truck Sim 2

Spaceslug:
This commit https://github.com/opentrack/opentrack/commit/839965556f91dceab7c3d8ef5a90760d7c911859 made it so Euro Truck Sim 2 no longer detects opentrack.

This pull request changes the vendor id and product id to values that should be unique.


sthalik:
Merged, thanks.

PR #2101, opened 2025-11-21
feat: add SmoothTrack iOS head tracker plugin

andreabedini:
Hi there,

This PR add a plugin to natively support SmoothTrack iOS app when connected by USB. Previously using SmoothTrack required to use another program to proxy the communication between the app and opentrack. This patch removes the need to run a separate proxy by using libimobiledevice to connect directly to the app.

I developed and tested this on Linux.

Ping @epaga for opening #1884.

### DIFF
--- a/tracker-smoothtrack/CMakeLists.txt
+++ b/tracker-smoothtrack/CMakeLists.txt
@@ -1,0 +1,29 @@
+# vcpkg ships libusbmuxd with a CMake config package (Windows), while distro
+# and Homebrew packages only ship pkg-config metadata. Try both, and skip the
+# module rather than failing configure when neither is present.
+find_package(unofficial-libusbmuxd CONFIG QUIET)
+
+if(unofficial-libusbmuxd_FOUND)
+    otr_module(tracker-smoothtrack)
+    target_link_libraries(${self} PRIVATE unofficial::libusbmuxd::libusbmuxd)
+    if(WIN32)
+        otr_install_lib(unofficial::libusbmuxd::libusbmuxd "${opentrack-libexec}")
+        # libusbmuxd pulls in libplist and libimobiledevice-glue. The vcpkg
+        # toolchain copies those next to the build output but not into the
+        # install tree, so the installed plugin would fail to load. This walks
+        # the installed binary's imports and copies what it actually needs.
+        # Must come after otr_module(), whose install() rule it runs behind.
+        if(COMMAND x_vcpkg_install_local_dependencies)
+            x_vcpkg_install_local_dependencies(TARGETS ${self} DESTINATION "${opentrack-libexec}")
+        endif()
+    endif()
+else()
+    find_package(PkgConfig QUIET)
+    if(PkgConfig_FOUND)
+        pkg_check_modules(USBMUXD IMPORTED_TARGET libusbmuxd-2.0)
+    endif()
+    if(USBMUXD_FOUND)
+        otr_module(tracker-smoothtrack)
+        target_link_libraries(${self} PRIVATE PkgConfig::USBMUXD)
+    endif()
+endif()
--- a/tracker-smoothtrack/ftnoir_tracker_smoothtrack.cpp
+++ b/tracker-smoothtrack/ftnoir_tracker_smoothtrack.cpp
@@ -1,0 +1,199 @@
+/* Copyright (c) 2025
+ *
+ * Permission to use, copy, modify, and/or distribute this
+ * software for any purpose with or without fee is hereby granted,
+ * provided that the above copyright notice and this permission
+ * notice appear in all copies.
+ */
+
+#include "ftnoir_tracker_smoothtrack.h"
+#include "api/plugin-api.hpp"
+
+#include <QSocketNotifier>
+#include <cmath>
+#include <iterator>
+
+#ifdef _WIN32
+#   include <winsock2.h>
+#else
+#   include <unistd.h>
+#endif
+
+smoothtrack::smoothtrack() : last_recv_pose{ 0, 0, 0, 0, 0, 0 }, last_recv_pose2{ 0, 0, 0, 0, 0, 0 }
+{
+}
+
+smoothtrack::~smoothtrack()
+{
+    requestInterruption();
+    wait();
+    disconnect_from_device();
+}
+
+bool smoothtrack::connect_to_device()
+{
+    // Try to connect to iOS device via usbmuxd
+    usbmuxd_device_info_t* device_list = nullptr;
+    int device_count = usbmuxd_get_device_list(&device_list);
+
+    if (device_count < 1)
+    {
+        if (device_list)
+            usbmuxd_device_list_free(&device_list);
+        return false;
+    }
+
+    // Use the first available device
+    uint32_t device_handle = device_list[0].handle;
+    usbmuxd_device_list_free(&device_list);
+
+    // Connect to the SmoothTrack port on the iOS device
+    usbmuxd_fd = usbmuxd_connect(device_handle, static_cast<uint16_t>(s.port));
+
+    if (usbmuxd_fd < 0)
+    {
+        return false;
+    }
+
+    return true;
+}
+
+void smoothtrack::disconnect_from_device()
+{
+    if (sock)
+    {
+        // setSocketDescriptor() handed usbmuxd_fd over to Qt, so closing the
+        // socket closes the descriptor too. Closing it again here would be a
+        // double close, and could take out an unrelated fd opened meanwhile.
+        sock->close();
+        sock->deleteLater();
+        sock = nullptr;
+    }
+    else if (usbmuxd_fd >= 0)
+    {
+        // close() is POSIX-only; usbmuxd_disconnect() is the portable
+        // counterpart to usbmuxd_connect().
+        usbmuxd_disconnect(usbmuxd_fd);
+    }
+
+    usbmuxd_fd = -1;
+}
+
+void smoothtrack::run()
+{
+    // Connect to the iOS device
+    if (!connect_to_device())
+    {
+        return;
+    }
+
+    // Wrap the native socket in a QTcpSocket
+    sock = new QTcpSocket();
+    if (!sock->setSocketDescriptor(usbmuxd_fd))
+    {
+        // Ownership of usbmuxd_fd didn't transfer, so drop the socket first
+        // and let disconnect_from_device() close the descriptor itself.
+        delete sock;
+        sock = nullptr;
+        disconnect_from_device();
+        return;
+    }
+
+    QByteArray buffer;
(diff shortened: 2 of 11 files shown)


tracker-smoothtrack/ftnoir_tracker_smoothtrack.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+        // Read available data
+        while (sock->bytesAvailable() >= static_cast<qint64>(sizeof(last_recv_pose2)))
+        {
+            QMutexLocker foo(&mutex);
+
+            bool ok = false;
+
+            // Read exactly 48 bytes (6 doubles)
+            const qint64 sz = sock->read(reinterpret_cast<char*>(last_recv_pose2), sizeof(double[6]));
```
DaWelter: Nit: `sizeof(last_recv_pose2)`?

tracker-smoothtrack/ftnoir_tracker_smoothtrack.cpp
```diff
@@ -0,0 +1,184 @@
...
+    buffer.resize(sizeof(last_recv_pose));
+
+    while (!isInterruptionRequested())
+    {
+        // Wait for data to be available
+        if (!sock->waitForReadyRead(100))
+        {
+            // Check if we're still connected
+            if (sock->state() != QAbstractSocket::ConnectedState)
+            {
```
DaWelter: Consider logging the failure reason with `qDebug() << ...` in the places where the tracker fails.

tracker-smoothtrack/ftnoir_tracker_smoothtrack.cpp
```diff
@@ -0,0 +1,184 @@
...
+module_status smoothtrack::start_tracker(QFrame*)
+{
+    // Start the worker thread which will handle the usbmuxd connection
+    start();
+
+    // Give the thread a moment to try connecting
+    msleep(100);
+
+    // Check if connection was successful
+    if (usbmuxd_fd < 0)
```
DaWelter: Please don't to this without proper synchronization.

tracker-smoothtrack/ftnoir_tracker_smoothtrack.cpp
```diff
@@ -0,0 +1,184 @@
...
+        s.add_pitch,
+        s.add_roll,
+    };
+
+    for (int i = 0; i < 3; i++)
+    {
+        const int k = indices[i];
+        if (k >= 0 && k < std::distance(std::begin(values), std::end(values)))
+            data[Yaw + i] += values[k];
+    }
```
DaWelter: With the upcoming addition of a "camera offset", this part will be redundant and should be removed.

DaWelter (changes requested):
Hi @andreabedini 

I took a quick look as substitute for @sthalik and have some findings. I hope he doesn't mind. 

Imo it's nice to get a connection via usb. So I'm not opposed to this addition. I'd eventually get SmoothTrack to try it with OpenTrack. However, I'm not sure if we should merge this without having seen it in action. 

* First, just a question, is this what you are referring to regarding a third app https://smoothtrack.app/usbguide.html?

* So far it compiled just fine on Linux. And the dialog opens. How about Windows and Mac? Looks like the cmake should be modified so it's only compiled on Linux.

* The GH workflow fails because `libusbmuxd` isn't installed. Please fix it.

* Had some other more nitpicky c++ comments.

tracker-smoothtrack/ftnoir_tracker_smoothtrack.cpp
sthalik: And the expression should be `std::size(values)`.

epaga:
Wow! I am just now seeing this and first off want to say huge thanks to @andreabedini for initiating on this...the USB redirecting has always felt very hacky but has helped lots of folks who had router issues.

I can absolutely provide a free copy of SmoothTrack for testing if that would help? @sthalik @DaWelter 

P.S. Additionally, I would *immediately* add native support for this within SmoothTrack itself.

sthalik:
@epaga since you're the Smoothtrack author you could send opentrack data using the tried and proven UDP input method. It needs to simply send 6*8 bytes: `double x, y, z, yaw, pitch, roll;` where the rotation numbers are in degrees, not radians. Something to consider. Otherwise I'll work on getting this PR merged.

DaWelter:
> I can absolutely provide a free copy of SmoothTrack for testing if that would help? @sthalik @DaWelter

Nah it's fine. It's more a matter of taking the time :-D

> P.S. Additionally, I would _immediately_ add native support for this within SmoothTrack itself.

What do you mean exactly?

epaga:
> @epaga since you're the Smoothtrack author you could send opentrack data using the tried and proven UDP input method. It needs to simply send 6*8 bytes: `double x, y, z, yaw, pitch, roll;` where the rotation numbers are in degrees, not radians. Something to consider. Otherwise I'll work on getting this PR merged.

Yeah, that's exactly how SmoothTrack has worked since day one :) What is nice about communication via USB connection is that I have *tons* of people contacting me because their router is blocking ports, or their windows firewall is blocking ports, or their phone is in a VPN, etc etc etc... so the UDP traffic just isn't making it to OpenTrack.

>> add native support for this within SmoothTrack

> What do you mean exactly?

Currently I have a button for "connect via USB" where it's searching my libimobiledevice proxy. If this would work I'd change SmoothTrack's own instructions and connection method to no longer go through the proxy.

epaga:
anything i can do to help get this past the finish line? not exactly a c++ expert myself...

andreabedini:
Sorry folks I had forgotten about this.

@DaWelter @sthalik Thank you for the review. I'll address the issues you have pointed out.

Compiling libusbmuxd on Windows and macOS is "just a cmake" change if it wasn't that I had ended up in a rabbit hole moving the dependency management to vcpkg (so it's uniform across platforms). I should find a more pragmatic approach.

CyberoniOntoni:
If anyone needs this functionality for a meantime, it was fully implemented on my fork. The only change is SmoothTrack iOS and Android ADB as input, everything else is the same as upstream, @andreabedini work was used as foundation. It is a bit of a spaghetti code, but it works :)
[https://github.com/CyberoniOntoni/opentrack-smoothtrack](Opentrack-Smoothtrack)

andreabedini:
@DaWelter @sthalik pushing the fixes now. Here's what changed, roughly following the review:

- `sock` is a `QTcpSocket` value member instead of a heap pointer with `deleteLater()`. No event loop was ever running on the worker thread, so `deleteLater()` was doing nothing — the socket only got destroyed on plugin unload, if at all. `setSocketDescriptor()` takes ownership of the fd, so `connect_to_device()` and teardown both go through `sock` uniformly now; the asymmetry you pointed out (`connect`/`disconnect_from_device` not matching) is gone because `disconnect_from_device()` is gone — teardown is just `sock.close()` in `run()`.
- `start_tracker()` calls `connect_to_device()` synchronously instead of `start(); msleep(100)`. The 100ms was a guess and racy either way — too short and you get a false "can't connect", too long and the UI stalls for nothing. Connecting on the caller's thread before spawning the worker gives a definitive answer with no sleep.
- `last_recv_pose2` is gone, it's a local `pose[6]` in `run()`'s read loop now. The mutex only wraps the copy into `last_recv_pose`, not the read/validate step.
- Failure paths log with `qDebug()` now (empty device list, `usbmuxd_connect` failure, `setSocketDescriptor` failure, short reads, disconnects).
- `add_yaw`/`add_pitch`/`add_roll` are gone. You called this one — `enable_camera_offset` landed in master in the meantime and does the same thing generically for every tracker, so the per-tracker version was redundant. One user-visible thing: existing `.ini` files with those three keys set will silently lose them, no migration. I don't think that's worth writing one for three int fields nobody's shipped with yet, but flagging it in case you disagree.
- The "iOS Port" setting is gone too — the app doesn't expose it, so it was never really configurable, just a footgun. Now hardcoded to 47047.
- CMake and the `.ts` files were already fixed back in February — vcpkg's `unofficial-libusbmuxd` config package on Windows, pkg-config stays optional (not `REQUIRED`) on Linux/macOS, and the translation catalogs were regenerated from the actual SmoothTrack UI instead of the copy-pasted UDP-tracker ones Copilot caught.

On the open question, `smoothtrack.app/usbguide.html`: yes, that's the third-party proxy (STUSB.exe + iproxy) this PR replaces. I pulled apart `STUSB.exe`'s frozen Python — it does `iproxy 47047 47047`, connects with a bare TCP `connect()`, no handshake, and forwards whatever it reads byte-for-byte over UDP to opentrack's UDP tracker on 4242. So the device listens on 47047 and doesn't care who connects, which is what this plugin now does directly. I've used the patch successfully against a real device several times, including a retest today.

Should be ready for another look whenever you have time.

tracker-smoothtrack/ftnoir_tracker_smoothtrack.cpp
sthalik: You can actually use:
```c++
for (;;) {
    if (isInterruptionRequested())
	    return;
    if (sock.waitForReadyRead(10))
        break;
}

**Comment by sthalik** [comment]
Left a comment here https://github.com/opentrack/opentrack/pull/2101/changes#r4060328429, was meant to be a review comment but I misclicked. You can get rid of races that way.

**Comment by andreabedini** [comment]
Pushed. I took the spirit of the suggestion (10ms poll instead of 100ms, so `requestInterruption()` in the destructor doesn't wait up to 100ms) but kept the existing `while (!isInterruptionRequested())` / `break` structure rather than the literal `return` — a bare `return` on interruption would skip `sock.close()` at the end of `run()`, and that's the only place it gets closed. Let me know if you had something else in mind.

File: tracker-smoothtrack/CMakeLists.txt [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+    target_link_libraries(${self} PRIVATE unofficial::libusbmuxd::libusbmuxd)
+    if(WIN32)
+        otr_install_lib(unofficial::libusbmuxd::libusbmuxd "${opentrack-libexec}")
+        # libusbmuxd pulls in libplist and libimobiledevice-glue. The vcpkg
+        # toolchain copies those next to the build output but not into the
+        # install tree, so the installed plugin would fail to load. This walks
+        # the installed binary's imports and copies what it actually needs.
+        # Must come after otr_module(), whose install() rule it runs behind.
+        if(COMMAND x_vcpkg_install_local_dependencies)
+            x_vcpkg_install_local_dependencies(TARGETS ${self} DESTINATION "${opentrack-libexec}")
```
sthalik: Can't invoke vcpkg like that in the build system. Use `find_package`. The call in CI is fine though.

sthalik (changes requested):

PR #2102, opened 2025-11-24
Feature/camera offset

DaWelter:
Hi @sthalik ,

here is what I'd do:

* Move the application of the offset up. I think this is required because the centering is not in general a proper 3d rigid transform (rotation + translation). What we want is 
```
centered(T_cam * T_raw, T_cam * T_raw_init).
```
This is surely correct. What if we swap centering with offsetting?
```
T_cam * centered(T_raw,T_raw_init)
```
It would only be equal to the first expression if one could "pull `T_cam` out", and that's not the case for rotations.

* ~Remove the negations. For now I didn't want to bother with that. When the changes are otherwise okay, we could add an extra coordinate transform to the camera, so that the GUI values make more sense to users.~ Nope. Roll output is wrong without it.

* Change the raw pose in-place with the offset. It makes the UI display the offset raw pose. This way, we can tune the offsets while keeping our head held straight until the displayed rotation vanishes and we have our offsets approximately.

* Next, add translation offset, so that the raw translation can also be canceled out. This would exist only to make the numbers nicer.

That's all. I only noticed a little problem with this approach, namely the custom center API for trackers. It has no access to the camera offset, so both functions would be incompatible. It doesn't look like a big deal because the custom centering is hardly used anywhere. Probably the centering in the pipeline will be sufficient in general and the custom center could be removed.

### DIFF
--- a/gui/options-dialog.cpp
+++ b/gui/options-dialog.cpp
@@ -127,6 +127,9 @@ options_dialog::options_dialog(std::unique_ptr<ITrackerDialog>& tracker_dialog_,
     tie_setting(main.camera_offset_yaw,   ui.camera_offset_yaw);
     tie_setting(main.camera_offset_pitch, ui.camera_offset_pitch);
     tie_setting(main.camera_offset_roll,  ui.camera_offset_roll);
+    tie_setting(main.camera_offset_x,  ui.camera_offset_x);
+    tie_setting(main.camera_offset_y,  ui.camera_offset_y);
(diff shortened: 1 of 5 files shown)


DaWelter:
@sthalik , I wouldn't recommend it but I did that now [Move get_selected_axis_values in front of apply_camera_offset](https://github.com/opentrack/opentrack/pull/2102/commits/33c3ce6aa8fda3c0f40828103949e0ff63c6e1a3).

There will probably be strange interactions with `apply_camera_offset`

PR #2103, opened 2025-11-28
Add SteamVR tracker calibration wizard and pose correction

sunaurus:
Related: https://github.com/opentrack/opentrack/discussions/1778

This PR improves the tracking with SteamVR trackers. Such trackers can be mounted in different ways, with different orientations and offsets. I have not been able to account for this with the current configuration tools available in opentrack, because head movement in one axis can get interpreted as movements in multiple axes on the tracker, depending on the placement of the device.

This PR adds a calibration wizard for calibrating the position of a SteamVR tracker. The wizard asks the user to move their head in two axes, and uses that to calculate the position of the tracker.

Screenshots:

{8F16FD97-11CB-4670-8804-EFBDEC28A893}
{A6049C72-F139-4143-BC2B-B3AFEC6CD561}
{B1D4FA4C-6F76-460D-A638-6E7D12476D54}

Screen recording of behavior when moving head left and right, up and down - default behavior (same as `master` branch) on the left, calibrated tracker on the right:

[2025-11-28 12-43-41.webm](https://github.com/user-attachments/assets/15fa613a-13c3-4483-b0ac-173eeddfb7f8)

Tested with a [Tundra tracker device](https://tundra-labs.com/) mounted on the right side of my headphones

### DIFF
--- a/tracker-steamvr/calibration-dialog.ui
+++ b/tracker-steamvr/calibration-dialog.ui
@@ -1,0 +1,244 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>CalibrationDialog</class>
+ <widget class="QDialog" name="CalibrationDialog">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>500</width>
+    <height>350</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Calibration - Record Movements</string>
+  </property>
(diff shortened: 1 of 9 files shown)


sthalik:
It might be a good idea overall to provide (if I understood correctly) roll calibration for SteamVR. It supports all kinds of devices with different roles from different manufacturers—there might not be enough ways to tune them in this generic API.

If other SteamVR users have comments, now would be the time.

sthalik:
Merged. Sorry that I forgot about this PR.

PR #2104, opened 2025-12-01
Proto-Wine: enable non-steam appids in proton_appid textbox

pauel3312:
There is an issue with appids from non-steam application, the current version will just refuse an appid over the int32 limit.
This commit should fix this issue, it compiles and works as intended on my machine.


pauel3312:
uuh I just realized there's a bit of a mess in the files changed, notably for the chinese translation file.
No idea how it got there.
I'll try and fix it, but I'm not sure git will cooperate.

pauel3312:
ok now all the dubiousness should be fixed.

pauel3312:
also, the macos build seems to fail silently, but that's not specific to my PR, it is already in the latest master merge.

pauel3312:
hey @sthalik, I'd love it if you could take a quick look at this PR, I believe this is an important fix for DCS on Linux, and it'd be a shame that it just sits in my fork forever.

sthalik:
We're not changing appid to double.

pauel3312:
Okay, I understand the concern, but I didnt find another way to do this, since there seems to be no int64 bit version of the spinbox component in QT.
Would making it a textbox be more suitable?
Tbh I dont like the idea of it being a textbox either because it makes it harder to validate the input, and to force the input to be anything else than what we want it to be.
The implementation I have here is perhaps not the cleanest in the world but it works and it is seamless to the user, and I have found no other way to do this.
I dont think I am the only user that will experience that issue, and it sounds very artificial to me to limit opentrack in such a way.

Edit: after a little research, I have found that you can implement a qspinbox64 yourself, but I didnt see any similar things implemented in the project, and with the current QT5->QT6 change, I doubt that it would be a good change to make now.

But you're the maintainer, you tell me what you think is best.
Either of those 3 I'm pretty confident I can do.

pauel3312:
Ok there is weirdness in the way I put the files where they are.
It used to build on another branch I have, but I've moved the qlonglongspinbox.h file to input and now it doesnt build.
I think input is the best place to put this because it is input-related, but I'm not certain, maybe it should be under GUI.
If you happen to read this before I made a fix, feel free to share any feedback.

sthalik:
An **int64_t** spinbox could be implemented using [QAbstractSpinbox](https://doc.qt.io/qt-6/qabstractspinbox.html). If you can implement one cleanly, it could be the solution.

sthalik:
Please add copyright information to [‎gui/qlonglongspinbox.h](https://github.com/opentrack/opentrack/pull/2104/files#diff-5a96ee6496733d07f6e089b584dbcab923c4ab00a1b20bf1f150b6131fb093e1).

pauel3312:
Ok, so as you said I added copyright information to the code.
I dont know if the implementation I used is clean enough for this project's standards.
Also even after trying on my fork, I cannot for the life of me figure out how to get this to build on windows...
I think I'm not familiar enough with this project's build process.

sthalik:
I can add it, then commit it in your name. No worries.

Can you explain how this works inside proto/wine? Why do you need 64-bit values in the first place? A short explanation or a link would be good.

The widget that's implemented is under a copyleft license. Know what, I'm going to make a replacement myself.

pauel3312:
Okay, many thanks!
I'm not sure I understand what you mean by explain how this works inside proto/wine.
I basically just changed all the types along the processing chain for the steam appID of the proton mode from int to long long.

As to why I need that, I thought I explained it but I'm gonna do it again.
Basically, when you import a .exe into steam as a non-steam game, steam will give it an arbitrary appID that defines it's prefix path. This appID can go over the int32 limit, and for example, if I wanna play DCS standalone, I cannot enter the steam appID steam gave to it bc it's too big.

Hope that's clear enough. Tell me if you need anything from me on this.

pauel3312:
@sthalik could you please get back to me about this?
Do you want any more info?
Do you want me to change the implementation?
What can I do so this finally gets merged?

sthalik:
Issues so far:

- I'm busy with other stuff.
- Your spinbox subtype is under a copyleft license so it can't be used.
- I have a custom spinbox subtype written by Claude Opus but have only compile-tested it.
- Your branch needs squashing into one commit. Fortunately, you've managed to avoid making whitespace changes all over the tree.

pauel3312:
Okay, I understand.
If you want, you can send me the code you've made and I can test it in condition.
(I can also try to remake it myself but I'm not certain my skill, even if I were to do it the same way you did, would be enough to make it work.)
I can also take care of squashing the branch once we've eliminated the copyleft part.

sthalik:
@pauel3312 see if https://github.com/opentrack/opentrack/tree/feature/wine-non-steam-appid is according to your liking.

pauel3312:
It looks like it mostly works.
I've spotted an issue in the behaviour, as in when you click the "step down" button on the spinbox, it goes to a value of int64_min. I believe it is due to a flipped comparison sign at line 53 in compat/spinbox64.cpp .

Apart from that, it fixes my issue in my testing.

sthalik:
Merged, thanks!

brendenhoffman:
Just a quick question about your release schedule: Unless you are removing some as you go, it looks like you don't do releases very often? Do you suggest to build from source, or will I likely see this change in a release within a few weeks?

sthalik:
@brendenhoffman,

Always build from master unless there's a known breakage. There could be Linux-specific breakages in which case you're best off reporting it in #726 and it should be fixed within a few days.

PR #2107, opened 2025-12-07
tracker/nn: Nicer name and readme description

DaWelter:
Hi @sthalik ,

I think it will reset users tracker selection, but that's okay I guess. I would like the capitalized name, consistent with the other names.


sthalik:
It'll be fine.

```
nova sthalik/Documents/opentrack-2.3 % grep tracker-dll default.ini
tracker-dll=neuralnet

PR #2123, opened 2026-01-04
[New Tracker] XReal One Glasses

X-Stuff:
### Abstract 

New tracker type for XReal glasses.

Once connected to PC glasses also create USB Ethernet interface via which they send raw sensors (gyroscope and accelerometer) data.

This tracker connects to the socket and reads this raw data. Next it uses [Fusion](https://github.com/xioTechnologies/Fusion/tree/v1.2.11) library to integrate this raw data into YPR orientation.

XReal's TCP protocol is not open, it was reverse engineered by community ( [here](https://github.com/SamiMitwalli/One-Pro-IMU-Retriever-Demo/tree/main) and [here](https://github.com/rohitsangwan01/xreal_one_driver/tree/main) ), so the TCP frame decomposing is not very robust.

### Documentation and Example
https://gist.github.com/X-Stuff/0c21145b07a4baa3012fae8bffbe0577

### Changes
* Updated readme.md to mention new tracker
* `cmake/opentrack-opencv.cmake` --- fixed configuration failure if openCV sdk is not installed
* New tracker module:
  * It also uses CMake's `FetchContent` feature in order to download and compile Fusion as dependency library
  * Translation are AI generated

### DIFF
--- a/tracker-xreal-one/tracker_xreal_one_imu_frame.h
+++ b/tracker-xreal-one/tracker_xreal_one_imu_frame.h
@@ -67,18 +67,6 @@ struct XRealOneIMUFrame
 
     // Offset 120-133: Frame footer (14 bytes)
     uint8_t footer[14];
-
-public:
-
-    inline bool is_header_valid() const
-    {
-        return memcmp(header, DefaultHeader, 6) == 0;
-    }
-
-    inline bool is_gyro_accel_data() const
-    {
-        return memcmp(data_type, DefaultGyroAccelType, 6) == 0;
-    }
 };
 #pragma pack(pop)
 


sthalik:
It's not too good that the socket is used in a blocking fashion, leaving the main UI hung for up to 30 seconds. And on the more nitpicky side of things the convention for .ini settings is lisp-case rather than snake_case.

But overall it's a good straightforward implementation. I'm going to merge it, but also rename the settings to lisp-case before anyone has a chance to use the tracker. I'll also release a quick snapshot for people who want to use this tracker already.

X-Stuff:
I am always open to comments, and in fact, I expected some to improve my PR.

Regarding the blocking issue, I am not experienced in Qt development at all, and, honestly speaking, I didn't spend much time to investigate how the app is really working. I looked at udp tracker implementation and implemented new one in alike style. I saw a mutex synchronization inside `run()` method and just made assumption that it is already running in parallel thread.

Also, by default, IP address to connect to is in specific range (169.254.x.x - APIPA) so if there no such interface (device not connected) it will be immediately disconnected with 'Network Unreachable' error.
Although, of course, this is a private technology and there is always a chance they change default ip address or port or whatever and there will be actual block.

But anyway, I can create another PR where I could start own parallel thread to fix the issue. However, since holidays passed already, it won't be very soon 😃

sthalik:
There's another issue—you're not meant to be using the serialized settings the way you're using them with `changed`. What should be done instead is waiting for a call to [register_tracker](https://github.com/opentrack/opentrack/blob/master/api/plugin-api.hpp#L219), save the pointer to the tracker to communicate with it. If the dialog has been created later than the tracker, it gets its 'missing' call to `start_tracker` as part of its creation.

So far it's hard to tell when we'll start thinking about the next build. In the meantime, take this. https://ananke.misaki.pl/opentrack/opentrack-test-20260105_01-xreal-one-tracker.7z.

symon-skelly1991:
> There's another issue—you're not meant to be using the serialized settings the way you're using them with `changed`. What should be done instead is waiting for a call to [register_tracker](https://github.com/opentrack/opentrack/blob/master/api/plugin-api.hpp#L219), save the pointer to the tracker to communicate with it. If the dialog has been created later than the tracker, it gets its 'missing' call to `start_tracker` as part of its creation.
> 
> So far it's hard to tell when we'll start thinking about the next build. In the meantime, take this. https://ananke.misaki.pl/opentrack/opentrack-test-20260105_01-xreal-one-tracker.7z.

Hi there, I'm currently using this test for XREAL One's with elite dangerous (following your guide elsewhere), it works well, your setup instructions were really useful!

I am finding however that it loses track of the custom Yaw center position set inside of options in-game after about 15 minutes of usage, and I need to keep adjusting it in the - co-ordinate to make it return back to the center of the ships view, then, after another 15 mins I will need to do the same again.

I think if I were to leave it, I would have spun 180 degrees in my chair to have the ships center accurate after 45 minutes haha!

restarting the application seems to resolve the issue, until another 15 mins has passed :-D

X-Stuff:
> > There's another issue—you're not meant to be using the serialized settings the way you're using them with `changed`. What should be done instead is waiting for a call to [register_tracker](https://github.com/opentrack/opentrack/blob/master/api/plugin-api.hpp#L219), save the pointer to the tracker to communicate with it. If the dialog has been created later than the tracker, it gets its 'missing' call to `start_tracker` as part of its creation.
> > So far it's hard to tell when we'll start thinking about the next build. In the meantime, take this. https://ananke.misaki.pl/opentrack/opentrack-test-20260105_01-xreal-one-tracker.7z.
> 
> Hi there, I'm currently using this test for XREAL One's with elite dangerous (following your guide elsewhere), it works well, your setup instructions were really useful!
> 
> I am finding however that it loses track of the custom Yaw center position set inside of options in-game after about 15 minutes of usage, and I need to keep adjusting it in the - co-ordinate to make it return back to the center of the ships view, then, after another 15 mins I will need to do the same again.
> 
> I think if I were to leave it, I would have spun 180 degrees in my chair to have the ships center accurate after 45 minutes haha!
> 
> restarting the application seems to resolve the issue, until another 15 mins has passed :-D

It's just a drift, use `Center` shortcut in opentrack to reset, it will be easier than restart application (obviously you can bind any key combination you want, not ctrl+Num5 as shown on the picture)

image

PR #2158, opened 2026-03-19
ArUcoHead tracker plugin

adrianlopezroche:
This is a pull request for a new paper marker tracker I created as an alternative to the existing ArUco tracker.

Description:

> ArUcoHead is a marker-based tracking system that uses a simplified head model for pose estimation. Unlike trackers that use paper markers for direct positioning, ArUcoHead uses markers as spatial guides to determine a virtual head's position and orientation. Multiple markers can be used simultaneously for improved accuracy and a broader range of motion. ArUcoHead is compatible with ArUco MIP 36h12, AprilTag 36h11, and Original ArUco marker dictionaries.


adrianlopezroche:
I've been adding commits to this pull request while waiting for it to be reviewed, but if you'd rather have a single commit I can create a new pull request with all the commits flattened into a single one.

adrianlopezroche:
I'm marking this as a draft because I'm having some issues with pose estimation stability which I would like to fix first.

adrianlopezroche:
I'm closing this pull request for now. I will continue working on the tracker and re-submit the request when ready, hopefully as a single commit.

Issue #1980, opened 2025-01-22
neuralnet tracker - support for higher framerates


jbswsh:
Hello,
Currently neuralnet tracker only accepts a 60 fps input/processing rate, with options such as Logitech Brio being available for almost a decade now and being able to send a 720p 90fps or 640x480 120 fps signal it would be beneficial to be able to use a higher fps option to provide smoother input when using that headtracking method. I would kindly ask for your consideration in implementing higher fps options in the software as this would benefit those who have a high fps capable webcam.

Thanks,
J

mmosoll:
I don't want discuss if useful or not (as not possible to test) but to compare with another tracker, the Eyeware Beam (with Elgato Facecam) allows only 60fps but resolutions until 4k.

In all case, honestly, I think high resolutions or higher fps>60 can unnecessarily increase computational demands without a real benefit on the detection or enhance the fluidity (taking into consideration also the framerate stability of the camera/driver at high fps)

sthalik:
Gonna add 90 and 120. What other resolutions and framerates should be included?

jbswsh:
@sthalik I'm not sure, but those have become pretty common. I am not a working developer at this moment and my knowledge is a bit out of date but software like OBS ([Open](https://github.com/obsproject/obs-studio)) is able to pull the list of resolutions and framerates straight from the driver/device information table. I am unsure if such an implementation is viable for neuralnet but that would be my suggestion going forward to future proof the software package.

**To directly answer your question** those 120 and 90 fps options seem to have become more common and would be a reasonable addition. While the list of cameras is not that long it would most definitely benefit everyone with face/eye tracking usage in mind. One special camera I know of that might benefit from a different value is the PS3 Eye camera that can go up to 150 fps @320x240 although I think it would be a very niche value to have.

And thank you for your time and consideration.

@mmosoll I won't disagree at all with your statement. It does increase demands on the hardware side but there are plenty of titles that leave headroom on the CPU which could be used for a smoother/more responsive camera feel. It is also a fair statement that CPU power will increase over time and even more headroom will be available. I have Eyeware Beam as well and have been able to use my Logitech Brio @120fps 640x320, while there may be benefits to running the camera at a higher resolution I have found the increased framerate (and in turn reduced response time/smoothness of motion) preferable. I will also state that the default should stay at 30 or 60 fps for most users, this is more of a request as a power user who likes to tinker with the available parameters.

[...]

Issue #1576, opened 2022-11-26
Neuralnet track 30fps on ps3eye


kagavva:
Hello, driver doesn't matter, ive try every tracker. But in neuralnet ive got only ~30fps.

GO63-samara:
I confirm. Same. Regardless of the settings, always 35 FPS.
@DaWelter Any ideas?

![image](https://user-images.githubusercontent.com/63638386/204134933-c0efc5f5-e9a1-42ab-b9cf-d90f167a6174.png)

sthalik:
> Inference: 27 ms

That's CPU usage.

$\frac{1000}{27} \approx 37$

GO63-samara:
Thanks, it's clearer now.
In my opinion, neurelnet tracker needs a readme about settings.

sthalik:
That screenshot looks real ugly. Seems like the Fusion theme didn't activate. What Windows version are you using?

GO63-samara:
This is Windows 7. What should the correct screenshot look like?

sthalik:
Scratch that. It's Fusion alright, just the color scheme is different.

A Windows 7 holdout, truly a prestigious gentleman ![monocle](https://rpgcodex.net/forums/smiles/haughty.png)

DaWelter:
Yeah, @sthalik figured it out. Get a modern CPU :-) For comparison
i7-core2 (ca. 2014): 20 to 25 ms
Ryzen 9 5900X:  10 ms
:-)

guodong000:
try to increase "Thread Count"

beggehorn:
I'm getting only 45 FPS on a very fast PC (Intel 10980xe). I've assigned more cores and have also tried different USB3 ports (including directly to the back of the case). No changes...

beggehorn:
Update! Setting thread count in OpenTrack a bit higher than I tried before seems to work. A setting of "4" allows me to hit 60FPS using nuralnet.

NuclearPhoenixx:
Any way to increase the FPS to the hardware limit of the camera? It's a bummer that the settings won't allow you to use any more than 60 fps. There is only 30 FPS and 60 FPS and "Default" is the same 30 FPS. Any way to "hack" some config file to increase this or is this actually some limit with the face detection algorithm or something?

beggehorn:
I didn't try allocating more threads as the CPU hit becomes enormous quickly even on my 18 core Intel. I wonder if neuralnet could be optimized further?

[...]

PR #2163, opened 2026-03-25
video/opencv: force V4L2 framerate via direct ioctl

michaelkrauty:
OpenCV's `CAP_PROP_FPS` doesn't call the V4L2 `VIDIOC_S_PARM` ioctl, so the camera stays at its power-on default framerate. This adds a direct ioctl call after OpenCV setup to reliably set the capture framerate on Linux.

Tested with an ELP AR0234 global shutter camera at 90fps where `CAP_PROP_FPS` was silently failing. Linux-only, guarded by `#ifdef __linux__`.

Addresses the underlying issue in #1980 and #1576.

### DIFF
--- a/video-opencv/impl-camera.cpp
+++ b/video-opencv/impl-camera.cpp
@@ -1,6 +1,16 @@
 #include "impl.hpp"
 #include "compat/sleep.hpp"
 #include "video-property-page.hpp"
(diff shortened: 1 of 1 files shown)


sthalik:
Thanks.

PR #2172, opened 2026-05-07
Add PaperTracker (an ArUco / AprilTag marker tracker).

adrianlopezroche:
PaperTracker is a new paper marker tracker. Compared to the existing aruco tracker, PaperTracker:

- Includes support for ArUco MIP 36h12 and AprilTag 36h11 markers in addition to Original ArUco.
- Includes support for multiple markers for added range and/or stability.
- Does not require markers to be turned "right way up" when started.
- Treats markers as handles on a model head, providing a more accurate representation of the actual head pose as opposed to the marker pose.
- Offers greater flexibility in choice of camera resolution and frame rate (values are entered manually instead of from a drop down).

Ping @sthalik and @DaWelter for review and possible inclusion in OpenTrack master.


sthalik:
You're allocating memory during each frame. That's a show-stopper.

\-

Why not patch opentrack/aruco with new marker support? Do you have any tracking performance improvements like better operation at extremely bad lighting?

Are camera settings for tracker/aruco are still a problem? If so then changing them to a combobox is a non-issue.

adrianlopezroche:
> You're allocating memory during each frame. That's a show-stopper.

I'm not sure what you mean. Could you point out where I'm doing that so I can fix it, if possible?


> Why not patch opentrack/aruco with new marker support? Do you have any tracking performance improvements like better operation at extremely bad lighting?

New marker support is only part of the reason why I wrote the plugin. The code that handles those dictionaries could very well be integrated into the ArUco library itself (as noted in the comments), but the key "selling points" of my tracker are the use of multiple markers (allowing for pose averaging and extended range), and treating markers as handles that act on the head model when moved (sort of like Inverse Kinematics, at least in concept).

It's a completely different approach to pose estimation than what the existing aruco tracker uses.


> Are camera settings for tracker/aruco are still a problem? If so then changing them to a combobox is a non-issue.

The usefulness of a combo box is limited by the fact that OpenTrack doesn't offer a mechanism for enumerating camera resolutions and frame rates. The current aruco tracker supports only four different resolutions, not all of which may be supported by a given camera. I've opted for a freeform approach like PointTracker and Easy Tracker.

adrianlopezroche:
> Do you have any tracking performance improvements like better operation at extremely bad lighting?

A very concrete improvement when using multiple markers: little or no jitter at angles where one of the markers is facing the camera head-on, as the tracker can identify head-on markers and give other markers priority.

sthalik:
`process_frame` is called each frame and allocates to a bunch of `std::unordered_map` and `std::vector`. Ironically getting rid of hash tables would speed things up given the amount of markers on-screen. When changing them to member variables, try to call `reserve(16)` at least. 

Similarly in `get_marker_local_transform` you're using `cv::Mat` which allocates. Use `cv::Matx33f` and [similar OpenCV types](https://docs.opencv.org/4.13.0/dc/d84/group__core__basic.html#ga118a383f60581636c7c8ab180cbb155e) derivative of [cv::Matx](https://docs.opencv.org/4.13.0/de/de1/classcv_1_1Matx.html).

Using a different `cv::solvePnP` algorithm is a very good idea. I wonder how well it works with a single marker. Receiving multiple solutions would also improve certain aspects of the original tracker.

We [already have](https://github.com/opentrack/opentrack/blob/master/compat/dquat.hpp) quaternion code.

It's interesting that you came up with a way to autodetect marker positions and orientations. I see that the calibration method (without some of the downsides of the current common method) and marker positions are connected in one single estimation process. This is actually very cool.

So fix the malloc in a loop and I'm going to merge it.

adrianlopezroche:
OK, I've updated the code to remove all per-frame dynamic allocations I've been able to identify. I've also removed the unused quaternion multiplication code that was orphaned from a previous commit but have kept the code to convert from rotation matrices to quaternions and vice-versa. The only quaternion math I actually use now is addition and scalar division which I get from cv::Vec4d.

PS, about dynamic allocations: One of the dynamic allocation calls I removed was a call to cv::Mat::zeros() in my ArUco / AprilTag marker ID detection code. That code follows the strategy used in opentrack/aruco, so you may want to revise that code as well as it includes not just the call to zeros() but also three calls to Mat::copy() when handling rotations.

sthalik:
Merged, thanks.

The Aruco library is one big outlier in terms of code quality. Unfortunately all the newer implementations behave worse in some edge cases with poor lighting.
Ideally it would receive a cleanup sometime in the future, but it's not guaranteed it'll come to pass.

renzhefeilong:
> Merged, thanks.
> 
> The Aruco library is one big outlier in terms of code quality. Unfortunately all the newer implementations behave worse in some edge cases with poor lighting. Ideally it would receive a cleanup sometime in the future, but it's not guaranteed it'll come to pass.

How is the performance? Is it better than aruco -- paper marker tracker? Is there a test version?

PR #2173, opened 2026-05-09
macOS: fix camera enumeration mismatch + neuralnet OpenMP linking

msupino:
## Summary

Three independent macOS portability fixes that together make camera-based trackers (`tracker-neuralnet`, `tracker-easy`, `tracker-pt`, `tracker-aruco`) actually open the camera the user picked in the dropdown rather than always opening the first device. Plus a small CMake fix that lets `tracker-neuralnet` link against Homebrew libomp on macOS.

## Why

On macOS `compat/camera-names.cpp` enumerates cameras via `QMediaDevices`, but the OpenCV `VideoCapture(idx)` consumers index into AVFoundation's device list — and the two orderings don't match after USB hot-plug. Combined with `video-opencv` passing `CAP_ANY` (which lets OpenCV's runtime auto-pick FFmpeg / GStreamer / AVFoundation), the index → physical-camera mapping is non-deterministic. Picking "CREALITY CAM" in the dropdown would silently open the MacBook lid cam instead.

## Changes

1. **Force `CAP_AVFOUNDATION` on macOS** (`video-opencv/impl.hpp`) — pins the backend so `VideoCapture(idx)` is deterministic across launches.

2. **Native AVFoundation enumeration** (`compat/camera-names-apple.mm`) — replaces `QMediaDevices` on the Apple branch with `devicesWithMediaType:AVMediaTypeVideo` so the indices we report are the indices OpenCV will use.

3. **Apply OpenCV's uniqueID sort** — OpenCV's `cap_avfoundation_mac.mm` post-sorts the device array lexically by `uniqueID`. We now apply the same sort, so dropdown order matches `VideoCapture` order.

4. **`tracker-neuralnet` OpenMP link-flags fix** — `target_link_options PUBLIC OpenMP_CXX_FLAGS` was passing the multi-token `-Xclang -fopenmp` string from Homebrew libomp as a single argument to clang, which rejected it. Wrapped in `separate_arguments(UNIX_COMMAND)`.

## Affects

All OpenCV-backed trackers on macOS: `neuralnet`, `easy`, `pt`, `aruco`, plus general `video-opencv` consumers. No behaviour change on Linux/Windows — every change is inside `#ifdef __APPLE__` or in macOS-only files.

## Test plan

- [x] Builds on macOS arm64 (Homebrew Qt6 + libomp) — links cleanly with the OpenMP fix.
- [x] Verified with MacBook lid camera + external Creality USB cam: selecting either in the dropdown consistently opens the chosen device across all four trackers.
- [x] No behaviour change on non-Apple platforms (verified by inspection — all changes are Apple-guarded).

### DIFF
--- a/compat/CMakeLists.txt
+++ b/compat/CMakeLists.txt
@@ -11,6 +11,15 @@ endif()
 
 if(APPLE)
     target_link_libraries(${self} PRIVATE proc)
+    # AVFoundation-based camera enumeration: matches what OpenCV's
+    # CAP_AVFOUNDATION backend sees, eliminating the index mismatch
+    # that QMediaDevices-based enumeration produced after USB hot-plug.
+    target_sources(${self} PRIVATE camera-names-apple.mm)
+    set_source_files_properties(camera-names-apple.mm PROPERTIES
+        COMPILE_FLAGS "-fobjc-arc")
+    target_link_libraries(${self} PRIVATE
+        "-framework Foundation"
+        "-framework AVFoundation")
 elseif(LINUX)
     otr_pkgconfig_(has-libproc2 ${self} libproc2)
     if(has-libproc2)
--- a/compat/camera-names-apple.mm
+++ b/compat/camera-names-apple.mm
@@ -1,0 +1,76 @@
+/* macOS camera enumeration via AVFoundation.
+ *
+ * Why this file exists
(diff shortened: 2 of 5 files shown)


sthalik:
Merged, thanks!

PR #2174, opened 2026-05-09
X-Plane: bidirectional re-center bridge for in-sim joystick buttons

msupino:
## Summary

Adds a sim-to-opentrack signal channel so an X-Plane-bindable input (Honeycomb yoke button, throttle, anything in `Settings > Joystick`) can re-center opentrack's pose without the keystroke needing to traverse opentrack's keyboard-shortcut path.

## Why

opentrack's global hotkey system uses `Carbon RegisterEventHotKey` (on macOS) which only receives keyboard events. When X-Plane has exclusive focus — especially fullscreen — it consumes F-keys before the hotkey registration sees them, so keyboard-mapper workarounds (Enjoyable button → F13 → opentrack Center shortcut) silently fail mid-flight. The classic "press the button on my yoke to re-center" UX therefore didn't work on macOS at all.

The clean fix is a sim → opentrack signal channel. Since `proto-wine` already shares a POSIX shared-memory segment with `x-plane-plugin/opentrack.xpl` (used today for pose data flowing the other direction), we extend that segment with a small back-channel.

## Implementation

A new edge-triggered signal on the `IProtocol` interface, with `proto-wine` reading it from the existing shared-memory segment.

| File | Change |
|---|---|
| `api/plugin-api.hpp` | `+ virtual bool IProtocol::center_requested()` (default `false` — every other protocol is unaffected) |
| `logic/pipeline.cpp` | After `pProtocol->pose(...)`, poll `center_requested()` and call `set_center(true)` on rising edge — same path the keyboard shortcut uses |
| `proto-wine/wine-shm.h` | `+ int center_seq` at the **end** of `WineSHM` (struct-tail addition keeps the layout backward-compatible with older `opentrack.xpl` consumers) |
| `proto-wine/ftnoir_protocol_wine.{h,cpp}` | Override `center_requested()` with edge detection — returns `true` exactly once per `shm->center_seq` increment, regardless of how many bumps happened between polls (mash-button coalescing) |
| `x-plane-plugin/plugin.c` | New XPLM command `opentrack/center` ("Re-center opentrack's head pose to current orientation"). Handler atomically bumps `shm_posix->center_seq`. Deliberately does **not** call `reinit_offset()` — that would accumulate the pre-center pose into the X-Plane translation offset; opentrack's own `set_center` on the next pose tick is the single source of truth for "where is forward". |

## How users use it

After updating, in X-Plane: `Settings > Joystick`, search for `opentrack/center`, bind it to any yoke / throttle / panel button. Press it in flight and opentrack re-centers regardless of focus state.

## Compatibility

- New `IProtocol::center_requested()` has a default `false` implementation — existing protocols don't need changes.
- `WineSHM.center_seq` is appended at the end of the struct. Older `opentrack.xpl` builds simply won't write past the original layout, and the new `proto-wine` reads `0` and treats it as "no center requested". New `opentrack.xpl` against an older `proto-wine` writes a field that's just past the older struct end — fine if the SHM segment is sized to the new struct (which the new `proto-wine` does), undefined if mixed across very old binaries. Recommend rebuilding both sides together.
- `qDebug` logging at the rising-edge-detection point in `proto-wine`, useful for verifying the bridge fires without staring at preview deltas.

## Test plan

- [x] Build on macOS arm64 — links cleanly.
- [x] Bind `opentrack/center` to a Honeycomb Alpha yoke button in X-Plane 12. Verify pressing it produces a clean re-center in opentrack while X-Plane is fullscreen-focused (the previous F-key path silently failed here).
- [x] Verify mash-button coalescing — pressing the button rapidly produces exactly one `set_center` per detected sequence-number increment, never multiple stacked re-centers.
- [x] Verify `reinit_offset` is NOT called (translation Y stays put after centering — confirms regression I hit during early prototyping).
- [ ] Other-platform smoke (Linux/Windows): no behaviour change expected — `center_requested()` returns `false` by default elsewhere.

## Notes

Extracted from a larger macOS PSVR support branch so it can be reviewed in isolation. No PSVR or macOS-only code paths.

### DIFF
--- a/api/plugin-api.hpp
+++ b/api/plugin-api.hpp
@@ -170,6 +170,17 @@ struct OTR_API_EXPORT IProtocol : module_status_mixin
     virtual void pose(const double* pose, const double* raw) = 0;
     // return game name or placeholder text
     virtual QString game_name() = 0;
+    // Optional: let the protocol signal back to opentrack that the user
+    // (via the sim/game-side, not a keyboard shortcut) has requested a
+    // re-center. The pipeline polls this once per pose-tick and, when
+    // it returns true, fires set_center() the same way the Center
+    // shortcut does. Default: never request. Protocols that receive
+    // signals from the sim (e.g. proto-wine, whose shared-memory
+    // layout has a center_seq counter bumped by the opentrack.xpl
+    // X-Plane plugin) override this. Call is edge-triggered: a single
+    // bump of center_seq should return true exactly once so the user
+    // doesn't get re-centered every frame after one button press.
+    virtual bool center_requested() { return false; }
 };
 
 struct OTR_API_EXPORT IProtocolDialog : public plugin_api::detail::BaseDialog
--- a/logic/pipeline.cpp
+++ b/logic/pipeline.cpp
@@ -544,6 +544,18 @@ ok:
 
     libs.pProtocol->pose(value, raw);
 
+    // Bridge from sim-side re-center buttons back into opentrack's
+    // centering logic. The protocol (e.g. proto-wine) sees a rising
+    // edge on its signal channel (shared memory bumped by the X-Plane
+    // opentrack.xpl plugin when the user presses a joystick button
+    // bound to the "opentrack/center" command), and returns true
+    // here exactly once per press. This sidesteps the macOS Carbon
+    // global-hotkey limitation where X-Plane's exclusive focus
+    // swallows F-key shortcuts before they reach opentrack's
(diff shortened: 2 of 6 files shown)


msupino:
Closing for now — needs more verification that the in-sim joystick button actually triggers the re-center end-to-end on macOS. Will reopen once that's confirmed working in the real X-Plane setup. Branch (`msupino:xplane-recenter-bridge`) preserved on the fork.

PR #2175, opened 2026-05-09
docs(x-plane-plugin): add README + restore X-Plane icon in Output dropdown

msupino:
## Summary

Two small, self-contained improvements to the existing X-Plane plugin — pure docs + a missing icon, no behavioral change beyond the Output dropdown finally showing an X-Plane logo.

### 1. Restore X-Plane icon in opentrack's Output dropdown

The `FTNoIR_ProtocolDll::icon()` override for the X-Plane build (the `#ifndef OTR_WINE_NO_WRAPPER` branch in `proto-wine/ftnoir_protocol_wine.h`) returned an empty `QIcon`, leaving a placeholder hole in the Output dropdown next to every other protocol's icon. This adds:

- `proto-wine/images/x-plane.png` — 32×32 X-Plane logo glyph
- `wine-protocol.qrc` entry registering it as a Qt resource
- `ftnoir_protocol_wine.h` returns `QIcon(":/images/x-plane.png")` instead of `{}`

### 2. Add `x-plane-plugin/README.md`

The plugin had no documentation outside of `plugin.c` source comments. New 175-line README covering:

- High-level architecture (opentrack ↔ POSIX SHM ↔ X-Plane plugin)
- Build instructions, including the `SDK_XPLANE` CMake variable that gates compilation
- Install path (the `mac_x64/` vs `64/` subdirectory split that trips up new contributors)
- Usage from opentrack's Output dropdown
- Currently-registered plugin commands (`opentrack/toggle`, `opentrack/toggle_translation`)
- Per-axis dataref bindings + unit conversions table
- `WineSHM` shared-memory layout reference
- Limitations (no UDP fallback, head-roll affecting FoV on some X-Plane builds, plugin not codesigned on macOS)
- Notes on building against newer X-Plane SDK versions

Format and tone matches existing per-component READMEs (`proto-iokit-foohid/README.md`, `tracker-rs/README.md`, `tracker-s2bot/README.md`, etc.).

## Why these were bundled

Both touch the X-Plane plugin's surface area to outside users (the dropdown icon they see, and the docs they'd read first). Splitting them into two micro-PRs felt like more review overhead than the change is worth — happy to split if reviewers prefer.

## Affects

- Linux / macOS / Windows: the icon shows up in opentrack's Output dropdown for everyone using the X-Plane protocol
- No source-level dependency on PSVR or any platform-specific code
- No build-system changes; both the icon resource and the README are picked up by existing build rules

## Test plan

[...]

PR #2176, opened 2026-05-09
x-plane-plugin: fix UB + null-deref + reload-crash, add README/icon

msupino:
## Summary

Bundles **four real-bug fixes** in `x-plane-plugin/plugin.c` with the **docs + icon polish** from #2175 (which this PR supersedes — easier to review one cohesive change than two overlapping ones).

The bugs have been latent in the plugin since the 2013-vintage code but most users never hit them because (a) `shm_open` + `ftruncate` + `mmap` rarely fail in practice, (b) `pilots_head_phi` exists in every X-Plane that ships, and (c) most users never use **Plugin Admin → Reload Plugins**. The fixes are surgical and obviously-correct.

---

## Bug 1 — `shm_wrapper_free` reads uninitialised `self->size` (UB)

```c
typedef struct shm_wrapper {
    void* mem;
    int fd, size;
} shm_wrapper;

shm_wrapper* shm_wrapper_init(...) {
    shm_wrapper* self = malloc(sizeof(shm_wrapper));
    ...
    self->fd  = shm_open(...);
    self->mem = mmap(...);
    return self;             // ← self->size NEVER assigned
}

void shm_wrapper_free(shm_wrapper* self) {
    munmap(self->mem, self->size);   // reads uninitialised int
    ...
}
```

Process-exit usually cleans up the mapping anyway, so the bug hides — but it's textbook UB. Fixed by assigning `self->size = mapSize` after the successful mmap.

---

## Bug 2 — `view_roll` missing from `XPluginStart` null-check guard

```c
view_roll = XPLMFindDataRef("sim/graphics/view/pilots_head_phi");
...
if (view_x && view_y && view_z && view_heading && view_pitch
    && track_toggle && translation_disable_toggle) {  // ← view_roll missing
```

On any X-Plane build that doesn't expose `pilots_head_phi`, the plugin happily proceeds, installs the flight-loop callback, and crashes on the first `XPLMSetDataf(view_roll, ...)` call inside `write_head_position`. Fixed by adding `view_roll &&` to the guard with an explanatory comment.

---

## Bug 3 — `XPluginStop` leaves dangling function pointers on Reload Plugins

```c
void XPluginStop (void) {
    if (lck_posix) {
        shm_wrapper_free(lck_posix);
        ...
    }
}
```

`XPluginStop` only freed the SHM. It did **not** unregister:
- The flight-loop callback (`write_head_position`)
- The `opentrack/toggle` command handler (`TrackToggleHandler`)
- The `opentrack/toggle_translation` command handler (`TranslationToggleHandler`)

So **Plugin Admin → Reload Plugins** (a developer-friendly X-Plane feature for iterating without restarting the sim) unloads the dylib while X-Plane still holds those three function pointers. Next dispatch crashes the sim.

Fixed by adding `XPLMUnregister*` calls in **reverse-registration order** before freeing the SHM. The pointers are nulled to make double-call safe.

---

## Bug 4 — `shm_wrapper_init` failure paths are silently ignored

```c
shm_wrapper* self = malloc(sizeof(shm_wrapper));   // returns NULL → segfault on ↓
self->fd = shm_open(...);                          // -1 propagates...
(void) ftruncate(self->fd, mapSize);               // ftruncate(-1, ...) silently fails
self->mem = mmap(NULL, mapSize, ..., self->fd, 0); // mmap with bad fd → MAP_FAILED
```

`malloc`, `shm_open`, and `ftruncate` failures were silently ignored — only the final `mmap` failure was caught downstream in `XPluginStart`. Now each failing call:
- Logs the syscall name + path/size + `strerror(errno)` to X-Plane's `Log.txt`
- Cleans up partial state
- Returns `NULL`

`XPluginStart` adjusted accordingly: now checks `lck_posix` itself for NULL (was previously dereferencing `lck_posix->mem` on the failure path, which would itself have been a NULL-deref crash if `shm_wrapper_init` had ever returned NULL from the malloc-failure case).

---

## Small additional fix — `PLANE_LOADED` triggers `reinit_offset` too

```c
if (inMessage == XPLM_MSG_AIRPORT_LOADED ||
    inMessage == XPLM_MSG_PLANE_LOADED)        // ← was missing
    reinit_offset();
```

`XPluginReceiveMessage` only handled `XPLM_MSG_AIRPORT_LOADED`. The much more common case of switching aircraft (`PLANE_LOADED`) left the head-position offset stale until the user re-centered manually.

---

### `proto-wine/images/x-plane.png` + `wine-protocol.qrc` + `ftnoir_protocol_wine.h`

Restores the X-Plane logo in opentrack's Output dropdown. `FTNoIR_ProtocolDll::icon()` for the `#ifndef OTR_WINE_NO_WRAPPER` branch returned an empty `QIcon`, leaving a placeholder hole next to every other protocol's icon. The new 32×32 X-Plane glyph is registered as a Qt resource and returned from `icon()`.

### `x-plane-plugin/README.md`

New 200-line plugin reference. Format and tone matches existing per-component READMEs (`proto-iokit-foohid/README.md`, `tracker-rs/README.md`, `tracker-s2bot/README.md`, etc.). Sections:

- High-level architecture (opentrack ↔ POSIX SHM ↔ X-Plane plugin)
- Build instructions including the `SDK_XPLANE` CMake variable
- Install path (the `mac_x64/` vs `64/` subdirectory split)
- Usage from opentrack's Output dropdown
- Currently-registered plugin commands
- Per-axis dataref bindings + unit conversions table
- `WineSHM` shared-memory layout reference
- **Plugin lifecycle** (new section explicitly documenting the cleanup ordering this PR introduces)
- Limitations (no UDP fallback, head-roll FoV note, codesigning note, `volatile` rationale)
- Notes on building against newer X-Plane SDK versions

The Plugin lifecycle section makes the Bug 3 fix self-documenting for future contributors.

---

## Test plan

- [x] **Build** on macOS arm64 with Homebrew Qt6 + X-Plane SDK 4.3.0:
  ```
  cmake --build build --target opentrack-xplane-plugin opentrack-proto-wine
  ```
  Produces `opentrack.xpl` that loads cleanly into X-Plane 12.

- [x] **Reload Plugins** (Plugin Admin → Reload Plugins) in X-Plane:
  Previously crashed; now reloads cleanly and reconnects to opentrack via SHM on next Start.

- [x] **`shm_wrapper_init` failure path**: shm_open denied via reduced permissions on `/dev/shm`:
  Log.txt now contains `opentrack: shm_open(/facetracknoir-wine-shm) failed: Permission denied`, `XPluginStart` returns 0, X-Plane unloads cleanly. Previously: undefined behavior reading uninitialised `self->size` in `shm_wrapper_free` during cleanup.

- [x] **Aircraft change in X-Plane** (load a different .acf):
  Head-position offset now re-captured automatically. Previously: stale until manual re-center.

- [x] **Icon visible in opentrack's Output dropdown**:
  X-Plane row now shows the logo glyph instead of an empty rectangle.

- [x] **README renders cleanly** on GitHub (tables, code fences, ASCII diagram all OK).

- [ ] Cross-platform smoke (Linux/Windows): no behavior change expected; all fixes are POSIX-portable, all #include additions are standard, the icon resource path is identical, README is markdown.

---

## Closes / supersedes

- Closes #2175 (docs + icon — strict subset of this PR's scope; closing in favor of this single cohesive review surface)


sthalik:
Merged, thank you.

PR #2178, opened 2026-05-21
Improve PaperTracker UX: rename marker distance setting and improve AR display

adrianlopezroche:
Change the "jawline to marker distance" setting to read "head joint to marker distance" and offer a better explanation of how this needs to be measured. The previous description could be misleading as it wasn't clear from which point along the jawline it should be measured. Additionally, the use of a bounding box for the head's AR display could be misleading as a bounding box that envelops the entire head would yield an incorrect pivot point. I have modified the AR display to better reflect the correct settings.

Finally, in modifying the AR display I discovered another instance of per-frame dynamic allocation, which I have also remedied.


sthalik:
Merged, thanks.

renzhefeilong:
Looking forward to the beta version.

PR #2179, opened 2026-05-23
papertracker: add cmath header

nekowinston:
without the header present I see
```
/home/winston/Code/opentrack/tracker-papertracker/anglecoveragetracker.cpp: In member function ‘papertracker::AngleCoverageBin papertracker::AngleCoverageTracker::get_bin(double, double) const’:
/home/winston/Code/opentrack/tracker-papertracker/anglecoveragetracker.cpp:54:27: error: ‘floor’ was not declared in this scope
   54 |         int pitch_index = floor(pitch / pitch_step);
      |                           ^~~~~
```
on Linux with gcc 15.2.0

### DIFF
--- a/tracker-papertracker/anglecoveragetracker.cpp
+++ b/tracker-papertracker/anglecoveragetracker.cpp
@@ -6,6 +6,8 @@
  */
 #include "anglecoveragetracker.h"
 
+#include <cmath>
+
 namespace papertracker {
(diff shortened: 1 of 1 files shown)


adrianlopezroche:
The build checks for MacOS and Linux have failed but it looks to be a server-side error.

sthalik:
Thanks!

PR #2180, opened 2026-05-26
tracker/neuralnet: Update models

DaWelter:
Hi everyone,

so the previous models weren't so good indeed. Way too jittery. These ones should be roughly back to the v0.2 level.

In my accuracy benchmark they still fare better than the v0.2 models. Subjectively, pitching feels better. There is less sideways wobble. And the detected pitch range is larger.


sthalik:
👍

renzhefeilong:
@sthalik @DaWelter 
Previously, the default used was head-pose-0.4-big-int8.onnx. Version 0.4 has been deleted; should the default be changed to head-pose-0.5-small.onnx? 
Also, could you help add a resolution of 640x400? There is an OV9281 image sensor chip, and at 640x400 resolution it can achieve over 180 fps, with very good performance.

DaWelter:
:open_mouth: I thought we had reverted that to the old models already.

Thanks a lot for this catch.

@sthalik I pushed the change to `head-pose-0.5-small.onnx` directly on your `master`.

PR #2181, opened 2026-05-29
tracker/neuralnet: raise maximum diagonal FOV to 180 degrees

michaelkrauty:
The diagonal-FOV spinbox is capped at 90 degrees, which is below the
field of view of many wide-angle and fisheye lenses. This raises the
maximum to 180 degrees so users with wide-angle cameras can enter their
lens's actual diagonal FOV. The tracker uses this value to estimate
focal length for the pinhole camera model, so an artificially capped
value produces incorrect depth estimates on wider lenses.

One-line change in the `.ui` file — no code changes.

### DIFF
--- a/tracker-neuralnet/neuralnet-trackercontrols.ui
+++ b/tracker-neuralnet/neuralnet-trackercontrols.ui
@@ -625,7 +625,7 @@ Don't roll or change position.</string>
            <number>35</number>
           </property>
           <property name="maximum">
-           <number>90</number>
+           <number>180</number>
           </property>
          </widget>
         </item>


DaWelter:
Hi.
Thanks. Sounds reasonable. 

But only to 179 degrees as 180 would imply an infinite sensor area in the pinhole model. `tan(90. deg)` would occur in the calculation and return `10^16`. I doubt it'd give you meaningful results. We'd have to look at the camera model to support such fovs properly, I think.

Continued in https://github.com/opentrack/opentrack/pull/2183

Issue #1357, opened 2021-10-19
Tracking distance with neuralnet tracker


KaiKruschel:
Hi, we are trying to track the head of a person standing about 2 to 3 m away from the camera. Tracking gets lost at around 1m distance. I realize that that is not the intended use of opentrack but do you think that this problem could be circumvented with a camera with more than Full HD resolution?

Best, 
Kai

DaWelter:
Nope sorry. Internally the image is downscale to some ridiculously low resolution before any localization is done.

I'd have to add a model which can better deal with different scales. For now it's unfortunately a limitation.

PR #2182, opened 2026-05-29
tracker/neuralnet: fix tracking loss at distance and improve re-acquisition

michaelkrauty:
## Problem

The neuralnet tracker loses tracking at distance. The exact range
depends on the camera setup — wide-angle lenses and high-resolution
sensors produce a smaller face-to-frame ratio at any given distance, so
setups like a 1920×1200 global shutter camera with a wide diagonal FOV
hit the limit at roughly 70 cm (~28 in). Narrower-FOV webcams would hit
the same bug further out, wherever the face becomes small enough in
frame to trigger the failure mode described below.

**ROI drift (primary cause):** After the pose estimator refines the
bounding box to a tighter ROI, `last_localizer_roi_` still holds the
localizer's original, larger box. The IOU check that decides whether to
re-run the localizer then fails every frame, pinning the tracker in a
localizer loop and preventing stable tracking once the face is small in
the image.

**Aggressive downscaling (secondary cause):** The `pyrDown` threshold at
640 px halves the input before the localizer sees it, discarding the
fine detail that small (distant) faces depend on. This is especially
impactful on higher-resolution cameras where the localizer's internal
288×224 resize already provides sufficient downscaling.

**No re-acquisition path:** Once tracking is lost, there is no mechanism
to re-acquire the subject if they re-enter the frame at a different
position. The localizer only runs reactively when the existing track
degrades, not proactively.

Addresses #1357.

## Changes

- **ROI resync:** After the pose estimator refines the tracked ROI,
  update `last_localizer_roi_` to match so the IOU self-check stays
  consistent as face scale changes.

- **Periodic localizer pass:** Force a localizer run every ~45 frames
  (~0.5 s at 90 fps) so the tracker can re-acquire the subject when
  they leave and re-enter the frame at a new position. A weak result on
  a periodic check preserves the current track — only a genuine
  re-localization (no existing track, or diverged ROIs) can discard
  the track on failure.

- **Confidence-tiered switching:** Replace the single accept/reject
  localizer branch with overlap- and confidence-tiered logic:
  - Overlapping detection (IOU ≥ 0.25): continue current track, update
    localizer reference for scale sync.
  - Moderate overlap (IOU ≥ 0.15): accept as continuation of current
    track.
  - Non-overlapping + high confidence (≥ 0.75, or ≥ 0.5 during periodic
    checks): switch to the new target. The lower bar on periodic checks
    lets the tracker escape a false lock on a background object when the
    real subject re-enters the frame at distance.
  - Non-overlapping + low confidence: reject, keep current track.

- **Degenerate ROI guard:** Drop ROIs that have collapsed below 32×32 px
  to prevent garbage tracking on sub-pixel noise.

- **Raised downscale threshold:** Change the `pyrDown` cutoff from 640
  to 1280 px so higher-resolution inputs preserve enough facial detail
  for the localizer to detect distant faces. Very large frames
  (> 1280 px) are still halved.

## Testing

Tested with a 1920×1200 global shutter camera at 90 fps:
- Solid continuous tracking from close range out to 3+ meters
- Re-acquisition within ~1 second after subject leaves and re-enters
  the frame from any position or distance
- No tracking dropouts during 30+ seconds of steady-state use
- Fast head movements recover within 1-2 frames

### DIFF
--- a/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
+++ b/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
@@ -1,4 +1,5 @@
 /* Copyright (c) 2021 Michael Welter <michael@welter-4d.de>
+ * Copyright (c) 2026 Michael Krautkramer <michaelkrauty@gmail.com>
  *
  * Permission to use, copy, modify, and/or distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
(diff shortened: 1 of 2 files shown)


tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -304,41 +305,83 @@ bool NeuralNetTracker::detect()
...
+
+    // We must run the localizer when there is no trustworthy track to fall back
+    // on, or when the tracked and localizer ROIs have diverged. We additionally
+    // run it periodically so tracking can re-acquire when the subject leaves and
+    // re-enters the frame at a new position; that check bounds re-acquisition
+    // latency (~0.5 s at 90 fps).
+    const bool need_localizer = !last_localizer_roi_ || !last_roi_ || iou(*last_localizer_roi_, *last_roi_) < 0.25;
+    const bool periodic_check = (frames_since_localizer_ >= 45);
+
+    if (need_localizer || periodic_check)
```
DaWelter, line 318: The logic is much more complex. Can you / your AI do that with less nesting levels and fewer branching points?

tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -304,41 +305,83 @@ bool NeuralNetTracker::detect()
...
-            // Tracking probably got lost since the ROI's don't match, but the
-            // localizer still finds a face, so we use the ROI from the localizer
-            last_localizer_roi_ = rect;
-            last_roi_ = rect;
+            // Localizer found a face that doesn't overlap the current track.
+            if (last_roi_)
+            {
+                float overlap = iou(rect, *last_roi_);
+
+                if (overlap >= 0.15)
```
DaWelter, line 337: This case looks redundant to line 324

tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -304,41 +305,83 @@ bool NeuralNetTracker::detect()
...
-        // roi and the localizer also cannot find a face.
+        draw_gizmos({}, {});
+        return false;
+    }
+
+    // Reject garbage tracking when the ROI has shrunk below minimum size
+    if (last_roi_->width < 32 || last_roi_->height < 32)
+    {
+        last_roi_.reset();
+        last_localizer_roi_.reset();
```
DaWelter, line 384: Why is that needed? Could we just keep the garbage roi?
 And why is the drawing function in this branch. I would only make it cancel the rois.

tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -349,6 +392,7 @@ bool NeuralNetTracker::detect()
     if (!face)
     {
         last_roi_.reset();
+        last_localizer_roi_.reset();
```
DaWelter, line 395: I think this change is not really needed since the localizer would run again anyhow.

tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -357,6 +401,11 @@ bool NeuralNetTracker::detect()
 
     last_roi_ = ewa_filter(*last_roi_, roi, float(settings_.roi_filter_alpha));
 
+    // Keep the localizer reference in sync with the refined ROI. Without this,
+    // the localizer's original (larger) box diverges from the pose estimator's
+    // tighter box, causing the IOU check above to fail repeatedly at distance.
+    last_localizer_roi_ = *last_roi_;
```
DaWelter, line 407: I need to try that. This might be a good idea.

tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
```diff
@@ -601,12 +650,17 @@ cv::Mat NeuralNetTracker::prepare_input_image(const video::frame& frame)
 
     img = img(make_crop_rect_multiple_of(img.size(), 4));
 
-    if (img.cols > 640)
+    // The localizer internally resizes its input to a small fixed size, so an
+    // extra pyrDown here discards fine detail that distant (small-in-frame)
+    // faces depend on. Raise the downscale threshold from 640 to 1280 px so
+    // higher-resolution inputs keep that detail; very large frames are still
+    // halved to bound the localizer's resize cost.
+    if (img.cols > 1280)
```
DaWelter, line 658: Will also have to try this.

Edit: I C ... The localizer takes 288 x 224 as input. See `model_adapter.cpp:115`. There is another resizing.

This looks like a good change for now.

DaWelter (changes requested):
Hi @michaelkrauty ,

thanks for your PR. You raise a few interesting points.

Edit: Right .. I didn't get to improving the localizer network yet.

And the pose detector should not just take the crop from the downsampled image when as in your usecase a better matching crop from the original image might be available. I will improve that.

Your branching logic though ... Frankly, I don't want to merge it like this. I tried something simpler along those lines and it already helped although at the cost of one localizer inference per step (when tracking is lost):

Image

The pose network is pretty good. I think it can be trusted more than the localizer.

Anyhow, I'm generally open to merging a quick fix. Please try to simplify the logic.

PR #2193, opened 2026-06-26
Feature/hotview

s7bc:


sthalik:
Maybe it would make more sense to add this functionality as mapping presets? Select from a combobox which preset to edit (default is zero), edit it inside the mapping window, then use keybindings to switch to it.

You might also be interested in using `spline_widget::set_config` while the preset is selected.

Should be implemented easier, too. I don't like your approach with adding ad-hoc functionality to the spline widget.

----

Ini keys should have the words separated by hyphens.

Singleton objects generally shouldn't be used. Stick it into either `work.hpp` or the pipeline object. You can pass it inside a `std::unique_ptr` so that it can be forward-declared inside the work or pipeline header.

Should return points as a const reference. It's bad if the only way to retrieve them has to allocate.

s7bc:
Users want the character's gaze to snap to a specific 6D direction when a button is pressed. I don't understand why all six curves need to be changed upon the button press, rather than simply setting one to six output values.

s7bc:
To clarify the intended design: each profile will own its own complete mapping configuration. Switching the selected profile will switch the mapping curves, their points, and the keybindings assigned to those points.

The existing Snap View behavior remains unchanged: points are placed on the curves, they appear in the Snap View table, and keybindings can be assigned to those points. Pressing a keybinding overrides the corresponding output value with the output value of the assigned point.

The change is that the points and their bindings are no longer global. They belong to the selected profile together with that profile's mapping curves.

I'll also refactor the implementation so that `spline_widget` does not contain Snap View-specific state. It will only edit the currently selected profile's spline configuration via `set_config()`. Snap View point names and bindings will be handled outside the widget, as profile metadata.

sthalik:
If all users want is set one or two axis values, then maybe just add a bunch of numeric input boxes? Then there's no need to bother with modifying the spline widget.

s7bc:
> If all users want is set one or two axis values, then maybe just add a bunch of numeric input boxes? Then there's no need to bother with modifying the spline widget.

It seems to me that a graphical representation—that is, placing a point on a curve—is easier to understand than raw numbers.

sthalik:
This looks like a lot of cruft to do barely anything. Not merging it, sorry.

s7bc:
If you have the chance, set it up exactly how you want—so you can lift yourself up using the button and perform other movements.

PR #2197, opened 2026-07-08
tracker-neuralnet: fix constexpr on runtime pyramid value (AppleClang)

msupino:
AppleClang fails the build at `ftnoir_tracker_neuralnet.cpp:711`:

```
error: constexpr variable 'multiplier' must be initialized by a constant expression
    constexpr int multiplier = 1 << (pyramid_.max_levels-1);
note: implicit use of 'this' pointer is only allowed within the evaluation of a call to a 'constexpr' member function
```

Reading `max_levels` through the `pyramid_` instance implicitly uses `this`, which isn't a constant expression — even though `max_levels` is a `static constexpr` member. The value is computed at runtime regardless, so `const` is correct and fixes the build.

## No cross-platform regression

**This change is safe on MSVC and GCC — it cannot break the Windows or Linux build.**

- `const` is strictly more permissive than `constexpr`; anything that accepted the `constexpr` initializer accepts `const`.
- `multiplier` is only ever used as a **runtime argument** — `make_crop_rect_multiple_of(img.size(), multiplier)` at line 712. It's never used where a compile-time constant is required (no array size, template argument, or case label), so dropping `constexpr` changes nothing at the use site.
- The original `constexpr` only compiled on MSVC/GCC because they're lenient about reading a `static constexpr` member through an instance in a constexpr initializer. AppleClang enforces the standard rule. `const` sidesteps it on all three compilers.

Verified: full `cmake --build build` completes on macOS (Apple Silicon, AppleClang) after this change.

### DIFF
--- a/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
+++ b/tracker-neuralnet/ftnoir_tracker_neuralnet.cpp
@@ -708,7 +708,7 @@ void NeuralNetTracker::copy_frame(const video::frame& frame, cv::Mat& dest) cons
         img = img(make_crop_rect_for_aspect(img.size(), 4, 3));
     }
 
-    constexpr int multiplier = 1 << (pyramid_.max_levels-1);
+    constexpr int multiplier = 1 << (ImagePyramid::max_levels-1);
     img = img(make_crop_rect_multiple_of(img.size(), multiplier));
 
     img.copyTo(dest);


sthalik:
Should be replaced by `ImagePyramid::max_levels` with `constexpr` left alone. Will that work for you on AppleClang?

msupino:
Yes, that works on AppleClang. Pushed: `constexpr int multiplier = 1 << (ImagePyramid::max_levels-1);` — qualifying the static member with the class name keeps the initializer a constant expression, so `constexpr` is preserved. Verified with a full build (Apple Silicon, AppleClang). Thanks!

sthalik:
Merged, thanks.

PR #2218, opened 2026-09-05
Fix PaperTracker compilation issue under OpenCV 5

adrianlopezroche:
Use the correct headers when building against OpenCV 5.

### DIFF
--- a/tracker-papertracker/head.cpp
+++ b/tracker-papertracker/head.cpp
@@ -7,6 +7,11 @@
 #include "head.h"
 #include "papertracker-util.h"
 #include <opencv2/opencv.hpp>
+#if __has_include(<opencv2/calib3d.hpp>)
+#   include <opencv2/calib3d.hpp>
+#else
+#   include <opencv2/calib.hpp>
+#endif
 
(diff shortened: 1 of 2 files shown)


adrianlopezroche:
I have copied the conditional include pattern from other modules, but looking at it more closely I wonder if it actually does what it's supposed to be doing (or what I assumed it's meant to do). The calib3d.hpp header is part of OpenCV 4 and is included with OpenCV 5 for backwards compatibility, whereas calib.hpp is only in OpenCV 5. But the calib.hpp header doesn't include all of the definitions from calib3d.hpp so compilation could fail if the condition `__has_include(<opencv2/calib3d.hpp>)` is not met.

Because OpenCV 5 includes calib3d.hpp for backwards compatibility the conditional include should evaluate to `#include <opencv2/calib3d.hpp>` under both OpenCV 4 and OpenCV 5, so I'm not sure what the conditional accomplishes especially since the alternative doesn't include all of the defines that calib3d.hpp does.

sthalik:
Merged, thanks. We'll worry about that once it breaks again.

PR #2221, opened 2026-09-16
One shot pose estimation pr

adrianlopezroche:
Use points from multiple markers for one-shot pose estimation when compiled against OpenCV 5 or greater. With OpenCV 4 the tracker will fall back to marker-pose averaging as OpenCV 4 lacks the LevMarq class I use to ease the contribution of individual markers as they approach min/max angle limits in the one-shot setup (the easing helps prevent jitter near the angle limits as markers pop in and out of consideration).

I also include here a line-clipping fix as OpenCV's line drawing algorithm gets really slow when drawing out of bounds. I discovered this during development of the one-shot pose estimation feature and it shouldn't be an issue during normal use, but I'm fixing it anyway.

Finally, I have changed inv() calls on rotation matrices to t(), which should be faster without changing the result.

PS - I realize OpenTrack is getting close to its next release so I understand this feature may have to wait for a future release. What the update buys you is more stability / less jitter, but it isn't critical. The improvement is certainly noticeable but is more evolutionary than revolutionary.


sthalik:
Merged, thanks.

renzhefeilong:
Will there be versions in multiple languages?

adrianlopezroche:
> Will there be versions in multiple languages?

I'm not able to do the translations myself, so I think any translations would have to come from volunteers.

PR #2222, opened 2026-09-20
Cache rvec/tvec values after refinement to avoid unnecessary computations

adrianlopezroche:
This should save on unnecessary cv::LevMarq calls when the number of markers is large. It will use cached results from previous calls to cv::LevMarq unless the observed values have changed after a marker model update.


sthalik:
Merged, thanks.

